PR #10, opened 2020-05-13
Draft RFC to add `Stream` to the standard library

nikomatsakis:


rfc-drafts/stream.md
```diff
@@ -0,0 +1,186 @@
...
+The reason that we have chosen to exclude combinators is that a number
+of them would require access to async closures. As of this writing,
+async closures are unstable and there are a number of [outstanding
+design issues] to be resolved before they are added. Therefore, we've
+decided to enable progress on the stream trait by stabilizing a core,
+and to come back to the problem of extending it with combinators.
+
+[outstanding design issues]: https://rust-lang.github.io/wg-async-foundations/design_notes/async_closures.html
+
+This path does carry some risk. Adding combinator methods can cause
```
sfackler, line 116: Assuming the futures crate is going to start reexporting std's Stream type when built against a suitable rust version, the ambiguity issue is going to exist regardless of when we add helper methods.

rfc-drafts/stream.md
```diff
@@ -0,0 +1,186 @@
...
+[summary]: #summary
+
+Introduce the `Stream` trait into the standard library, using the
+design from `futures`. Redirect the futures-stream definition to the
+standard library.
+
+# Motivation
+[motivation]: #motivation
+
+XXX describe async streams
```
cramertj, line 16: I'd expect to see this motivation section laying out the case specifically for having this trait in the standard library, rather than in futures-core. Some points to include are the perceived increase in stability, the increase in authority, and a promise of support from the language. It may also be worth discussing `stream fn`/`async gen fn` as a goal, as language items like these are one technical reason that would force this trait to be in the standard library (though the timetables for this + the uncertain API for synchronous generators and generator-backed iterators + concerns around "attached"ness make this argument fail to resonate with me at the moment).

rfc-drafts/stream.md
nikomatsakis: I'm not sure I follow. You're saying that if the futures crate does a 

```rust
pub use std::stream::Stream
```

or whatever, then this will trigger ambiguity?

Obviously it all depends on what mechanism we imagine. The most obvious is something like tagging the method as `#[low_priority]`, in which case we will prefer extension methods to the inherent method. It doesn't seem like it would matter then when we make the change.

Thinking more bout what I wrote, though, I was imagining that maybe we could do something like tag methods as `#[edition(2021)]` such that, in code before the edition, they are not visible, but I'm not sure what we would do *after* the edition. I guess that we could rewrite code to use `<Foo as StreamExt>::flat_map` or whatever but...nobody wants that.

Another option might be to *tag the `StreamExt` methods* so that they explicitly "cede" priority to std -- i.e., if we work in collaboration with itertools, futures, etc then people can seamlessly migrate to std.

rfc-drafts/stream.md
nikomatsakis: Yes, this section definitely needs to be elaborated. That's a good list of the points we should be hitting.

nikomatsakis:
Something else I was wondering: where in the standard library hierarchy should this trait live? I have to check what @sfackler did in the PR I guess. 

More generally I feel like we have to decide whether to just sprinkle async things here and there with names like `AsyncFoo` or to take some other approach (e.g., `std::async::...`, although that's a keyword). I guess maybe it depends on how many additions we ultimately hvae.

rfc-drafts/stream.md
sfackler: Oh sorry yeah, I was talking about risk in the absence of a new mechanism like `#[low_priority]`.

nikomatsakis: OK -- but even without that mechanism, if we never move methods over, we don't have a problem, right? It's only a problem once we start to move methods from the extension trait into the regular trait.

nikomatsakis: Ah, but we *will* have problems if we move some methods over now.

nikomatsakis: (But presumably we would remove them from `StreamExt` as part of the minor update.)

sfackler: Would code that invokes StreamExt methods via UFCS still work if the methods are moved to a super-trait?

yoshuawuyts:
> I guess maybe it depends on how many additions we ultimately have.

`async-std` is effectively a big experiment to see how much of the stdlib would be useful (and possible) to port to async Rust. And so far the it seems virtually everything `async-std` provides is seeing use by people. What this means for the stdlib is that it would be reasonable to leave ourselves with enough space to grow, even if we're currently only considering a modest set of APIs for inclusion. Examples of _"We may want to include a lot of this in the future"_ are: the `io` submodule (37 top-level exports) and `sync` submodule (18 top-level exports).

---

> where in the standard library hierarchy should this trait live?

That's a really good question. @sfackler added it under `core::stream`, which makes sense as a counterpart to `core::future`. But when we consider `AsyncRead` and `AsyncWrite` for inclusion there may be different tradeoffs and the obvious `std::io` submodule may not be the best fit.

rfc-drafts/stream.md
```diff
@@ -0,0 +1,186 @@
...
+resolution. We have had problems in the past with attempting to migate
+iterator helper methods from `itertools` for this same reason.
+
+While such breakage is technically permitted by our semver guidelines,
+it would obviously be best to avoid it, or at least to go to great
+lengths to mitigate its effects. One option would be to extend the
+language to allow method resolution to "favor" the extension trait in
+existing code, perhaps as part of an edition migration.
+
+## "Attached" streams
```
yoshuawuyts, line 127: It'd be great if this RFC could also mention async iteration syntax at least somewhere. I believe it's one of the requirements for "attached streams" to work so maybe there. Or perhaps it's noteworthy enough to mention under "future possibilities" by itself. It seems important to me to at least mention when discussing streams (:

rfc-drafts/stream.md
nikomatsakis: I think so, but it's worth testing

nikomatsakis: Yes [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=916f8bea3af6af3c7f75c28fdbcc3f37)

nikomatsakis:
I'm going to merge this as a WIP -- we'll need to do follow-up work to flesh it out.

Issue #12, opened 2020-05-26
Stream RFC


tmandry:
As drafted in #10

nellshamrell:
Doing research on this now :)

yoshuawuyts:
I was working on a crate today that models the browsers History API, and found that the concept of a `DoubleEndedStream` would be a great tool to model it: the history API is asynchronous and can can be navigated by going forward and backward through "the history stack". Under "future possibilities" it's probably worth mentioning we should explore not only `IntoStream` and `FromStream` which have clear tie-ins to language features. But we should also look towards filling out the roster of traits we found useful for `Iterator`.

Under [`async_std::stream`](https://docs.rs/async-std/1.6.0/async_std/stream/index.html) we've shown it's possible to create async counterparts to the traits in `std::iter`. I'm not sure if it's worth mentioning every one of these, but we've successfully been able to implement the following traits:
```txt
DoubleEndedStream  # A stream able to yield elements from both ends.
ExactSizeStream    # A stream that knows its exact length.
Extend             # Extends a collection with the contents of a stream.
FromStream         # Conversion from a Stream.
FusedStream        # A stream that always continues to yield None when exhausted.
IntoStream         # Conversion into a Stream.
Product            # Trait to represent types that can be created by multiplying the elements of a stream.
Stream             # An asynchronous stream of values.
Sum                # Trait to represent types that can be created by summing up a stream.
```

nellshamrell:
Got quite a bit more in place with #13.

To do on the next iteration:

- [ ] More information about generators
- [ ] Sketch out what IntoStream trait would look like
- [ ] Sketch out what FromStream trait would look like

nikomatsakis:
Some notes about attached vs detached streams and how interconversion might work (and what hurdles we know of, mostly a question of coherence) can be found in this hackmd:

https://hackmd.io/gHqlQBTiSrexwb-ppRPn-A

Discussion from [this point on Zulip](https://zulip-archive.rust-lang.org/187312wgasyncfoundations/77561addingStreamtostdlib.html#200261119)

nikomatsakis:
Some concerns that should block opening the RFC

* [ ] (Minor) Resolve the name question of attached vs lending
* [ ] Add `next` method to the stream so it can be used for something
* [ ] Add impls for `Box<T>` and other fundamental types
[...]

PR #13, opened 2020-06-03
[DRAFT] Update stream rfc

nellshamrell:
This is still a WIP.

[Rendered](https://github.com/nellshamrell/wg-async-foundations/blob/update-stream-rfc/rfc-drafts/stream.md)

### DIFF
--- a/rfc-drafts/stream.md
+++ b/rfc-drafts/stream.md
@@ -13,9 +13,16 @@ design from `futures`. Redirect the `Stream` trait definition in the
 # Motivation
 [motivation]: #motivation
 
-Streams are a core async abstraction. We want to enable portable libraries that produce/consume streams without being tied to a particular executor.
+Streams are a core async abstraction. We want to enable portable libraries that 
+produce/consume streams without being tied to a particular executor.
 
-People can do this currently using the `Stream` trait defined in the [futures](https://crates.io/crates/futures) crate. However, the stability guarantee of that trait would be clearer if it were added to the standard library. For example, if [Tokio](https://tokio.rs/) wishes to declare a [5 year stability period](http://smallcultfollowing.com/babysteps/blog/2020/02/11/async-interview-6-eliza-weisman/#communicating-stability), having the stream trait in std means there are no concerns about trait changing during that time ([citation](http://smallcultfollowing.com/babysteps/blog/2019/12/23/async-interview-3-carl-lerche/#what-should-we-do-next-stabilize-stream)).
+People can do this currently using the `Stream` trait defined in the 
+[futures](https://crates.io/crates/futures) crate. However, the 
+stability guarantee of that trait would be clearer if it were added 
+to the standard library. For example, if [Tokio](https://tokio.rs/) 
+wishes to declare a [5 year stability period](http://smallcultfollowing.com/babysteps/blog/2020/02/11/async-interview-6-eliza-weisman/#communicating-stability), 
+having the stream trait in the standard library means there are no concerns 
+about the trait changing during that time ([citation](http://smallcultfollowing.com/babysteps/blog/2019/12/23/async-interview-3-carl-lerche/#what-should-we-do-next-stabilize-stream)).
 
 ## Examples of crates that are consuming streams
 
@@ -126,6 +133,9 @@ why they ended up the way they did.
 
 `Stream` will live in the `core::stream` module and be re-exported as `std::stream`.
 
+It is possible that it could live in another area as well, though this followes
+the pattern of `core::future`.
+
 ## Why use a `poll` method?
 
 An alternative design for the stream trait would be to have a trait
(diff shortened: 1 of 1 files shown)


rfc-drafts/stream.md
```diff
@@ -233,19 +247,62 @@ Designing such a migration feature is out of scope for this RFC.
...
+
+[Source](https://smallcultfollowing.com/babysteps/blog/2019/12/10/async-interview-2-cramertj-part-2/#the-need-for-streaming-streams-and-iterators)
+
+In a **detached** stream, the `Item` that gets returned by `Stream` is "detached" from self. This means it can be stored and moved about independently from `self`.
+
+In an **attached** stream, the `Item` that gets returned by `Stream` may be borrowed from `self`. It can only be used as long as the `self` reference remains live.
+
+### Alternate terminology
+
+An alternative terminology - which may be easier to understand in terms of ownership and borrowed - involves **owned** and **borrowed** streams.
```
nikomatsakis, line 276: Probably we should settle on one terminology before publishing the RFC...I don't like the term "owned" tream because the question is not so much about *ownership* as *who is the owner* -- i.e., is it the stream or the consumer of the stream? And that is not clarified.

nellshamrell: True - I'm wondering if attached/detached might be the best terminology. It is not perfect, but it seems like the clearest.

nellshamrell:
@nikomatsakis thank you for the comments - I believe I have addressed all of them.

I also believe the Generator syntax section should be expanded. I would love some ideas of what to focus on.

rfc-drafts/stream.md
nikomatsakis: I guess one advantage of attached/detached is that it works both ways -- i.e., an "attached" stream means that the *stream and the items it producs are attached to one another*.

rfc-drafts/stream.md
```diff
@@ -221,31 +231,142 @@ existing code, perhaps as part of an edition migration.
 
 Designing such a migration feature is out of scope for this RFC.
 
-## IntoStream / FromStream traits, mirroring iterators
+## IntoStream / FromStream traits
+
+The `IntoStream` trait would provide a way to convert something into a `Stream`.
```
nikomatsakis, line 236: Probably good to elaborate a bit here and to mention that: Iterators have an `IntoIterator` trait that is used with `for` loops. We might want such a trait for streams, and it would look like ...

rfc-drafts/stream.md
```diff
@@ -221,31 +231,142 @@ existing code, perhaps as part of an edition migration.
...
+
+The `FromStream` trait would provide a way to convert a `Stream` into something else.
+
+Implementing the `IntoStream` and `FromStream` traits on `Stream` is currently blocked on needing [`async fn` available in traits]( http://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard).
```
nikomatsakis, line 240: ...that would then clarify why this is true (in fact, I'm not entirely convinced this is true)

rfc-drafts/stream.md
```diff
@@ -221,31 +231,142 @@ existing code, perhaps as part of an edition migration.
...
+## Other Traits
 
-* currently blocked on async fn in traits
-* The exact bounds are unclear.
-* the same as combinators
-* These would be needed to provide similar iteration semantics as Iterator:
-    * `for x in iter` uses `impl IntoIterator for T`
-    * `for x in &iter` uses `impl IntoIterator for &T`
-    * `for x in &mut iter` uses `impl IntoIterator for &mut T`
+Eventually, we will also want to add some (if not all) of the roster of traits we found useful for `Iterator`.
```
nikomatsakis, line 251: I guess I would weaken this to "we may also want" -- presumably the reason for wanting these "families" of traits is to enable combinator methods, so I might emphasize that. e.g., if we want to have `reverse` on streams, then we need `DoubleEndedStream`, etc

rfc-drafts/stream.md
```diff
@@ -221,31 +231,142 @@ existing code, perhaps as part of an edition migration.
...
+
+[async_std::stream](https://docs.rs/async-std/1.6.0/async_std/stream/index.html) has created several async counterparts to the traits in [std::iter](https://doc.rust-lang.org/std/iter/). These include:
+
+```
+DoubleEndedStream  # A stream able to yield elements from both ends.
```
nikomatsakis, line 256: can we make this a bullet list?

* `DoubleEndedStream` -- ...

nikomatsakis:
Oh, I guess there *is* a tracking issue. cc https://github.com/rust-lang/wg-async-foundations/issues/12, maybe you want to add some of the points and to-do items in there...

rfc-drafts/stream.md
nellshamrell: Very true!

nellshamrell:
Thank you @nikomatsakis! I've incorporated several of those suggestions and placed everything else in a checklist in #12.

Merging.

rfc-drafts/stream.md
```diff
@@ -233,19 +247,62 @@ Designing such a migration feature is out of scope for this RFC.
...
+```rust
+#[async]
+for elem in stream { ... }
+```
+
+Designing this extension is out of scope for this RFC.
+
+## "Attached" streams
+
+There has been much discussion around attached/detached streams.
```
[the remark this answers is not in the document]
tmandry: > every "Detached" steam can become an attached one, and _some_ detached streams can become attached ones, but that is harder to specify

I'm not quite sure I follow – I think you meant that some attached streams can become detached streams? In that case it seems clear that you would define it as `impl Stream` from the start, since that's more general?

PR #15, opened 2020-07-01
Update stream rfc

nellshamrell:


rfc-drafts/stream.md
```diff
@@ -124,6 +124,52 @@ where
...
+
+We should also implement a next method, similar to [the implementation in the futures crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12).
+
+```rust
+/// Future for the [`next`](super::StreamExt::next) method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct Next<'a, St: ?Sized> {
+    stream: &'a mut St,
+}
```
yoshuawuyts, line 137: Because this is the first-ever proposal for inclusion of a Stream adapter in the stdlib I think it may be useful to briefly discuss the name of the trait bounds. In the stdlib [generic `Iterator` bounds are referred to by `I`](https://doc.rust-lang.org/std/iter/struct.StepBy.html). Most bounds in the stdlib are single-character, so for consistency it would be better to name the bound `S` instead of `St`.

This is what we've done in `async-std` as well, and we haven't run into any conflicts with other bounds.
```suggestion
pub struct Next<'a, S: ?Sized> {
    stream: &'a mut S,
}
```

rfc-drafts/stream.md
```diff
@@ -124,6 +124,52 @@ where
 }
 ```
 
+## Next method/struct
+
+We should also implement a next method, similar to [the implementation in the futures crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12).
+
+```rust
+/// Future for the [`next`](super::StreamExt::next) method.
```
yoshuawuyts, line 132: This comment appears to be a copy-paste artifact. The `Iterator` adapters have a really nice comment scheme that I think we can adopt for `Stream` as well.

```suggestion
/// A future that advances the stream and returns the next value.
///
/// This `struct` is created by the [`next`] method on [`Stream`]. See its
/// documentation for more.
///
/// [`next`]: trait.Stream.html#method.next
/// [`Stream`]: trait.Stream.html
```

rfc-drafts/stream.md
```diff
@@ -124,6 +124,52 @@ where
...
+}
+
+impl<St: ?Sized + Stream + Unpin> Future for Next<'_, St> {
+    type Output = Option<St::Item>;
+
+    fn poll(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Self::Output> {
+        self.stream.poll_next_unpin(cx)
```
yoshuawuyts, line 160: `poll_next_unpin` isn't available in this proposal. Instead we can write:
```suggestion
        Pin::new(&mut *self.stream).poll_next(cx)
```

rfc-drafts/stream.md
```diff
@@ -235,27 +281,137 @@ Designing such a migration feature is out of scope for this RFC.
...
 This trait could look like this:
 
-[TO BE ADDED]
+```rust
+pub trait FromStream<A> {
+
+    fn from_stream<T>(iter: T) -> Self
+    where
+        T: IntoStream<Item = A>;
+}
```
yoshuawuyts, line 402: In [`async_std::stream::FromStream`](https://docs.rs/async-std/1.6.2/async_std/stream/trait.FromStream.html#tymethod.from_stream) the `from_stream` method emulates an `async fn`. This differs from `IntoStream` because:

- Going from `T -> Stream` can be sync because the stream itself is async.
- Going from `Stream -> T` (this trait) the yielded value *isn't* async, which means the method to create it needs to be.
```suggestion
pub trait FromStream<A> {
    async fn from_stream<T>(stream: T) -> Self
    where
        T: IntoStream<Item = A>;
}
```

rfc-drafts/stream.md
```diff
@@ -235,27 +281,137 @@ Designing such a migration feature is out of scope for this RFC.
...
+
+We could potentially include a collect method for Stream as well.
+
+```rust
+pub trait Stream {
+    fn collect<B>(self) -> B
+    where
+        B: FromStream<Self::Item>,
+    { ... }
+}
```
yoshuawuyts, line 413: Like `FromStream::from_stream` this method should be an `async fn` as well:

```suggestion
pub trait Stream {
    async fn collect<B>(self) -> B
    where
        B: FromStream<Self::Item>,
    { ... }
}
```

rfc-drafts/stream.md
```diff
@@ -289,25 +445,29 @@ for elem in stream { ... }
...
 [Source](https://smallcultfollowing.com/babysteps/blog/2019/12/10/async-interview-2-cramertj-part-2/#the-need-for-streaming-streams-and-iterators)
 
-In a **detached** stream, the `Item` that gets returned by `Stream` is "detached" from self. This means it can be stored and moved about independently from `self`.
 
-In an **attached** stream, the `Item` that gets returned by `Stream` may be borrowed from `self`. It can only be used as long as the `self` reference remains live.
+In an **lending** stream (also known as an "attached" stream), the `Item` that gets returned by `Stream` may be borrowed from `self`. It can only be used as long as the `self` reference remains live.
+
+In a **non-lending** stream (also known as a "detached" stream), the `Item` that gets returned by `Stream` is "detached" from self. This means it can be stored and moved about independently from `self`.
+
+This RFC does not cover the addition of lending streams (streams as implemented through this RFC are all non-lending streams).
```
yoshuawuyts, line 461: I really like this explanation!

rfc-drafts/stream.md
```diff
@@ -346,19 +506,42 @@ where
...
+
+and
+
+```rust
+impl<T> LendingStream for Box<T> where T: LendingStream
+```
+
+There is a coherence conflict for `Box<impl Stream>`, so presumably it will fail the coherence rules. 
+
+[More examples are available here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a667a7560f8dc97ab82a780e27dfc9eb).
```
yoshuawuyts, line 538: This is a really useful reference; with the context of the text in this RFC I feel like this is the best explainer I've encountered so far.

rfc-drafts/stream.md
yoshuawuyts: It's probably also worth noting that this family of traits may require a fair share of `Send` and `Sized` bounds ([ref](https://github.com/async-rs/async-std/pull/665)), but that's probably out of scope for this section.

rfc-drafts/stream.md
taiki-e: By the way, `futures-util` uses `St` for `Stream` because `S` may conflict with `Sink`. We use `S` in crate where `Sink` is not related, e.g., `futures-core`. It seems reasonable to use `S` here as `Sink` is not planned to be added to the standard library.

`futures-util` also use [`Fut` for `Future` because `F` may conflict with closures](https://github.com/rust-lang/futures-rs/blob/0.3.5/futures-util/src/future/future/map.rs#L13). `std` and other `futures-*` crates uses only `F`(this is related to that `std` has no utility like using both closures and futures). `async-std` uses both `Fut` and `F`, and it seems to be inconsistent overall (perhaps code directly copy-pasted from `futures-util` uses `Fut`?).

rfc-drafts/stream.md
taiki-e: By the way, originally `futures-util` used that style, but we changed it to the current simple style because:

* https://github.com/rust-lang/futures-rs/issues/1214:
  > DRY (do not repeat yourself): Let's _not_ include a short description. The combinator method always already has a more detailed, better-phrased description. Linking to that is enough/better.
* It needs to click the combinator type docs one by one to see which combinator method returns that type.

(I do not mean to strongly recommend the style of futures, but I don't think std's style is necessarily great.)

rfc-drafts/stream.md
```diff
@@ -235,27 +281,137 @@ Designing such a migration feature is out of scope for this RFC.
...
+pub trait IntoStream where 
+    <Self::IntoStream as Stream>::Item == Self::Item,
+{
+    type Item;
+
+    type IntoStream: Stream;
+
+    fn into_stream(self) -> Self::IntoStream;
+}
+```
```
taiki-e, line 350: One of the nice things about this trait is that it makes it very easy to write streams in combination with [async stream](https://github.com/taiki-e/futures-async-stream), like [IntoFuture](https://www.reddit.com/r/rust/comments/d7acl0/async_finalizers/f0zkpyg/).

```rust
type S(usize);

impl IntoStream for S {
    type Item = usize;
    type IntoStream: impl Stream<Item = Self::Item>;

    fn into_stream(self) -> Self::IntoStream {
        #[stream]
        async move {
            for i in 0..self.0 {
                yield i;
            }
        }
    }
}    
```

Currently, we need to add an inherent method that returns a stream and call that method, or write a manual poll_next method.

rfc-drafts/stream.md
nellshamrell: Thanks to @nikomatsakis!

skade:
[Rendered](https://github.com/nellshamrell/wg-async-foundations/blob/update-stream-rfc/rfc-drafts/stream.md)

rfc-drafts/stream.md
```diff
@@ -124,6 +124,52 @@ where
 }
 ```
 
+## Next method/struct
+
+We should also implement a next method, similar to [the implementation in the futures crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12).
```
nikomatsakis, line 129: I think when we give the definition of the trait earlier, we should include something like

```rust
// Defined in std::stream module
pub trait Stream {
    // Core items:
    type Item;
    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>>;
    
    // Optional optimization hint, just like with iterators:
    #[inline]
    fn size_hint(&self) -> (usize, Option<usize>) {
        (0, None)
    }

    // Convenience methods (covered later on in the RFC):
    fn next(&mut self) -> Next<'_, Self>
    where
        Self: Unpin;
}
```

nikomatsakis: I would also introduce this by saying "In general, we have purposefully kept the core trait definition minimal. There are a number of useful extension methods that are available, for example, in the futures-stream crate, but we have not included them because they involve closure arguments, and we have not yet finalized the design of async closures.

However, the core methods alone are extremely unergonomic. You can't even iterate over the items coming out of the stream. Therefore, we include a few minimal convenience methods that are not dependent on any unstable features. Most notably, `next`"

I guess I wrote this assuming we may add one or two more...

nellshamrell: Sounds good!

rfc-drafts/stream.md
```diff
@@ -373,9 +575,29 @@ or "owned" stream, the generator yield could return things
...
+After desugaring, this would result in a function like:
+
+```rust
+fn foo() -> impl Iterator<Item = Value>
+```
+
+```rust
+async gen fn foo() -> Value
+```
+
```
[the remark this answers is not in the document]
nellshamrell: Meant to have it in, but it's not obvious why. I will clarify :)

rfc-drafts/stream.md
```diff
@@ -235,27 +281,156 @@ Designing such a migration feature is out of scope for this RFC.
...
+**Streams**
+
 We may want a trait similar to this for `Stream`. The `FromStream` trait would provide way to convert a `Stream` into another type.
 
 This trait could look like this:
 
-[TO BE ADDED]
+```rust
+pub trait FromStream<A> {
+    async fn from_stream<T>(stream: T) -> Self
```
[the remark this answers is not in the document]
nellshamrell: I definitely think that's possible. To clarify, are you thinking of implementing this same trait for both Streams and Iterators, or are you thinking of having separate traits with very similar functionality?

nikomatsakis:
Reading the rendered form

image

I think those bullets were meant to be expanded upon -- and in fact I think we have, so they can just be removed.

(I can't figure out how to leave this as a comment on the diff, since they're not part of the diff...)

rfc-drafts/stream.md
```diff
@@ -63,14 +63,20 @@ the current task to be re-awoken when the data is ready.
...
+    // Optional optimization hint, just like with iterators:
     #[inline]
     fn size_hint(&self) -> (usize, Option<usize>) {
         (0, None)
     }
+
+    // Convenience methods (covered later on in the RFC):
+    fn next(&mut self) -> Next<'_, Self>
+    where
+        Self: Unpin;
```
nikomatsakis, line 79: can somebody (maybe @withoutboats or @cramertj) explain to me why `Self: Unpin` is here and why that's not a crushing limitation? I feel like I'm missing something super basic. I guess the answer is maybe that we expect to be operating with `Box<S>` where `S: Stream`, and `Box<S>` is always `Unpin`?

And in terms of why it is needed, I guess it is because: you can invoke `next` and that creates a `&mut` references to self, but once you've consumed the item, you want to be able to move `self` somewhere else .. I don't know. I can see *why* it requires `Unpin` (we are invoking `Pin::new` in the `next` method on `Next`), but I'm having a hard time with the "intuition" here.

I guess the alternative would be the `next` takes `Pin<&mut Self>` and not `&mut self`? Why would that be bad?

Anyway, I think we should have some text justifying it.

withoutboats: Sure, the reason why its needed is pretty clear if you look at the definition of Next:

``rust
struct Next<'a, S: Stream> {
    stream: &'a mut S
}
```

Since `Stream::poll_next` takes a pinned reference, the next future pretty clearly needs `S` to be `Unpin` to safely construct a `Pin<&mut S>` from `&mut S`. The alternative would be for next to take `Pin<&mut S>`.

You're basically right about operating on `Box<S>`, but remember that we can also operate on a stack-pinned `S` a lot of the time, so it doesn't necessarily require an allocation.

I think the current approach is strictly better than requiring the method to be pinned, because that would just require pinning it even when the type is unpin, whereas the current situation requires pinning it only when the type is unpin.

Right now, no one cares because we have no `!Unpin` streams in practice. But once we have async generators, that's going to be a big annoyance with using them. The problem is even worse with generator/Iterator, because generators won't implement Iterator if they have to be pinned to run, instead only pinned references to them will. I'm not sure what we're going to do about this when generators exist, this was one of the big papercuts we hadn't looked into yet.

File: rfc-drafts/stream.md (hunk 11 above)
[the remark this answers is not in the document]
Comment (tmandry) [reply rfc-drafts/stream.md:428]: I'm saying that `FromStream` (as defined here) and the existing `FromIterator` trait are separate, but with very similar functionality.

Thinking more about it, `FromStream` is actually more general, because an await point is allowed to suspend execution of the current function but doesn't actually have to. So many, but not all, existing impls of `FromIterator` would work for `FromStream` as well. If not for coherence with existing impls, we could do `impl<T: FromStream<Item = I>> FromIterator<I> for T`, and switch a bunch of `FromIterator` impls to be `FromStream` impls without having to repeat ourselves.

This is a somewhat similar problem to the `LendingStream`/`Stream` relationship below, but the relationship is reversed (the trait impls we have now are allowed to assume too much – that there are no suspend points – rather than too little).

This might be another possible bullet point for the lang team discussion, @nikomatsakis? (cc https://github.com/rust-lang/lang-team/issues/34)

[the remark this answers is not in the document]
Comment (yoshuawuyts) [reply rfc-drafts/stream.md:428]: The way we solved this in `async-std` is by manually copying all implementations of `FromIterator` from std [(ref)](https://docs.rs/async-std/1.6.2/async_std/stream/trait.FromStream.html). When we choose to introduce `std::stream::FromStream` we could pretty much copy-paste all of the impls. I forget whether there were coherence issues (likely), but the main reason why we took this approach is because doing general conversions between sync and async code is hard to get right in practice:
- a generic conversion from `sync -> async` means operations may quietly block in the background which can lead to perf issues in runtimes
- a generic conversion from `async -> sync` means using methods such as `task::block_on` which is hard to tune, or risk deadlocks when nested (both futures-rs and async-std have had issues with this)

To help with the conversion from `Iterator` -> `Stream` we introduced [`async_std::stream::from_iter`](https://docs.rs/async-std/1.6.2/async_std/stream/fn.from_iter.html). To my knowledge all of this together has covered people's needs quite well.

File: rfc-drafts/stream.md [hunk 13]
```diff
@@ -235,27 +291,156 @@ Designing such a migration feature is out of scope for this RFC.
...
+**Streams**
+
 We may want a trait similar to this for `Stream`. The `IntoStream` trait would provide a way to convert something into a `Stream`.
 
 This trait could look like this:
 
-[TO BE ADDED]
+```rust
+pub trait IntoStream where 
+    <Self::IntoStream as Stream>::Item == Self::Item,
```
shepmaster, line 352:
```suggestion
pub trait IntoStream
where 
    <Self::IntoStream as Stream>::Item == Self::Item,
```

rfc-drafts/stream.md
nikomatsakis: Yeah, so the problem will be that if `async gen` just returns an `impl Stream` (and not, crucially, an `impl Stream + Unpin`), you'll have to explicitly do some pinning operation to turn them into a "usable" stream (i.e., one you can call `next` one?).

The same seems to apply to a `gen fn`, but as you say, the situation is almost worse.

It does make me wonder if `Stream` should have a `poll_next(&mut self)` method instead of `poll_next(Pin<&mut self>)`, and we have the expectation that a `gen fn` should require some kind of "explicit operation" to make a `Stream` (which would basically be pinning it). The main benefit of this would be that `Iterator` kind of requires that. Basically, the argument to me for "let's just stabilize `Stream`" has relied in part on the idea that all the problems we face for Stream, we also face for Iterator, but actually in this situation the two designs diverge slightly (with Iterator potentially having "another problem", depending on your POV).

I guess that the idea would be that a `gen fn` actually doesn't return an `Iterator` but rather an `impl IntoIterator`, and an `async gen fn` returns an `impl IntoStream`, where the "into" part of it does the pinning required to make `&mut self` reasonable.

This seems like an important detail, I'm glad we're digging into it a bit more.

nikomatsakis: > Right now, no one cares because we have no `!Unpin` streams in practice.

To make sure I'm following, this is because people are hand-writing their streams, and basically here the `Pin`-machinery is just kind of boilerplate.

rfc-drafts/stream.md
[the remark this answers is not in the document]
taiki-e: I don't think the way currently used in `async-std` is the ideal way. That way means that all third-party crates that provide the `FromIterator` implementation must be patched in order to fully support.

As far as I know, this is why `futures` implements `collect` using existing traits (`Default`, `Extend`) rather than its own. The way used in `futures` cannot currently optimize based on the size-hint of the stream, but this will be possible once https://github.com/rust-lang/rust/issues/72631 stabilizes.

If we want the collection to extend asynchronously, you can use [`Sink`](https://docs.rs/futures/0.3.5/futures/sink/trait.Sink.html) or use a loop.

> To help with the conversion from Iterator -> Stream we introduced async_std::stream::from_iter. To my knowledge all of this together has covered people's needs quite well.

This seems copy of [futures::stream::iter](https://docs.rs/futures/0.3.5/futures/stream/fn.iter.html)?

rfc-drafts/stream.md
taiki-e: Here is an example when removing the `Unpin` bounds and using unsafe: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=526ee8e3ebca18a4a8756bff3984a23a) 
(this is unsound because you can move the underlying stream after `Next` dropped)

rfc-drafts/stream.md
```diff
@@ -124,6 +130,56 @@ where
 }
 ```
 
+## Next method/struct
+
+We should also implement a next method, similar to [the implementation in the futures crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12).
+
+In general, we have purposefully kept the core trait definition minimal. There are a number of useful extension methods that are available, for example, in the futures-stream crate, but we have not included them because they involve closure arguments, and we have not yet finalized the design of async closures.
```
taiki-e, line 137: > futures-stream crate

Did you mean futures-util or futures crate?

rfc-drafts/stream.md
withoutboats: > To make sure I'm following, this is because people are hand-writing their streams, and basically here the Pin-machinery is just kind of boilerplate.

Exactly. Hand-written streams/futures that are not `Unpin` are very rare. All the Pin machinery exists on the assumption that some day we will generate streams from functions and these will not implement Unpin.

withoutboats: > I guess that the idea would be that a gen fn actually doesn't return an Iterator but rather an impl IntoIterator, and an async gen fn returns an impl IntoStream, where the "into" part of it does the pinning required to make &mut self reasonable.

The only way to make a trait that does this is to box it, since you can't write a function that returns a stack pinned value.

Since the big use case for iterators (and probably for streams if we ever have a dedicated syntax) is to loop with `for`, which inherently "pins" them in place to the stack, we'll probably want something a bit more nuanced than them returning IntoIterator, so that for loops can somehow treat them as pinned without making an allocation (which in addition to its costs doesn't work at all in no-std no-alloc environments).

rfc-drafts/stream.md
nellshamrell: I meant future-utils, correcting :)

rfc-drafts/stream.md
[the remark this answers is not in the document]
yoshuawuyts: > If we want the collection to extend asynchronously, you can use Sink or use a loop.

I don't think that's a compelling outcome; this means that patterns from sync Rust wouldn't be transferable to async Rust. Where possible I think it's incredibly valuable to aim for parity between the two. Though this kind of gets into "design philosophy", and at least in the context of this PR we don't need to find consensus.

> That way means that all third-party crates that provide the `FromIterator` implementation must be patched in order to fully support.

That's accurate, and I think that's the right outcome. In some cases `FromIterator` implementations perform blocking IO which I think is I think enough reason to not introduce a blanket conversion. For example if a function were to use `impl FromStream<Result<fs::Metadata>>` as an input, then both an iterator constructed from `std::fs::ReadDir` and a stream constructed from `async_std::fs::ReadDir` would be valid*. However the synchronous variant would likely interfere with the scheduler.

*_Admittedly it's a bit of a contrived example, but hopefully it still serves to illustrate the point._

rfc-drafts/stream.md
```diff
@@ -372,10 +583,34 @@ yield could return references to local variables. Given a "detached"
 or "owned" stream, the generator yield could return things
 that you own or things that were borrowed from your caller.
 
+### In Iterators
+
 ```rust
-gen async fn foo() -> X {
+gen async fn foo() -> Value {
```
voidc, line 589:
```suggestion
gen fn foo() -> Value {
```

rfc-drafts/stream.md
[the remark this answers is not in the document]
taiki-e: > and at least in the context of this PR we don't need to find consensus.

Agreed, it actually requires real async trait method to add this to std, and afaik all of the existing approaches have some problems.

> In some cases `FromIterator` implementations perform blocking IO which I think is I think enough reason to not introduce a blanket conversion.

Well, this problem also exists in `futures::stream::iter`/`async_std::stream::from_iter`.

---

Aside from it's preferable or not, even a crate that is obviously unrelated to async can't be used as a return type for `collect` without implementing it, which is necessary information when explaining/considering-adding it (I mainly think about collections).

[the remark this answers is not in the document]
taiki-e: > Aside from it's preferable or not, even a crate that is obviously unrelated to async can't be used as a return type for `collect` without implementing it, which is necessary information when explaining/considering-adding it (I mainly think about collections).

One solution I was thinking about for this is, leave `futures`'s current `collect` method (as `collect2` or another name), even after the `FromStream` based `collect` is implemented.
(This is not an ideal solution, but it means third-party crates can provide mitigation for this, even if `std` can't solve this problem.)

rfc-drafts/stream.md
nellshamrell: Good catch, ty!

rfc-drafts/stream.md
nikomatsakis: OK, I've been turning this over. Bear with me.

I'm basically trying to justify a design where `Stream` requires a pinned self, even though `Iterator` does not, given that we want `Stream` and `Iterator` to work as analogously as possible. I guess my main concern is that `Iterator` doesn't require pinning in its core `next` method, so for a `gen fn` to interoperate with that ecosystem, there is going to have to be some kind of initial pinning step that converts its result into an iterator. This is going to be a bit tricky since, as you point out, you can't return a pinned value except by boxing, at least today. 

The point though is that the general shape will be `gen_fn().pin_somehow().adapter1().adapter2()`.

On the other hand, with streams, the core interface *is* pinned, and so the pinning occurs at the last moment, kind of right before you invoke `next`. Something like `async_gen_fn().adapter1().adapter2().pin_somehow()`.

"Pin at end" is probably the better setup, since it lets you build and return those adapters and then apply to pinning at the end. There are perhaps also efficiency gains from this setup, since maybe the adapters want to be pinned and can take advantage of that. So this is probably the argument in favor of the current design (and of course the analogy with futures).

I guess that implies that to have a `gen fn` that produces iterators will require *either* 

* pinning at the start, vs the end, which is asymmetric but maybe that's ok
* disallowing borrowing yields, also asymmetric
* some kind of `PinnedIterator` trait that can be "adapted" into an iterator (by pinning)
* some clever idea I've not had yet

Does that sound right?

tmandry: Just to make your third idea more concrete, for my own sake:

```rust
trait PinIterator {
  type Item;
  fn next(self: Pin<&mut Self>) -> Self::Item;
  // combinators can go here (duplicating Iterator for the most part)
}
impl<I: PinIterator, P: Deref<Target = I> + DerefMut> Iterator for Pin<P> {
  type Item = <I as PinIterator>::Item;
  fn next(&mut self) -> Self::Item { self.as_mut().next() }
}

// this would be nice.. but would lead to name resolution ambiguity for our combinators 😬 
default impl<T: Iterator> PinIterator for T { .. }
```

withoutboats: > I'm basically trying to justify a design where Stream requires a pinned self, even though Iterator does not, given that we want Stream and Iterator to work as analogously as possible.

The pinning also applies to the design of AsyncRead/AsyncWrite, which currently use Pin even though there's no clear plan to make them implemented with generator type syntax. In essence the asyncification of a singature is currently understood as pinned reciever + context arg + return poll.

> There are perhaps also efficiency gains from this setup, since maybe the adapters want to be pinned and can take advantage of that.

Pinning also introduces an indirection which will probably impact the optimizability of these adapters if you have to pin first.

> disallowing borrowing yields, also asymmetric

We should actually pursue this idea, it might work! Remember that we initially had only non-self-referential generators before adding self-referential generators that don't implement Unpin. It was immediately clear that holding a self-reference while doing nonblocking IO was super essential, but I'm not as certain that's true for iteration. It's possible that non-async generators could disallow self-references and just implement Unpin. That would certainly make the situation with Iterator much easier.

It's also possible that this would be a really bad limitation that makes them almost useless. But we need a real investigation to find out which it is.

nikomatsakis: With respect to generators that cannot borrow across a yield, I think it's not going to work. I imagine a lot of people are going to want to write code like this

```rust
fn consume(x: impl Iterator<Item = T>) { .. }

consume(gen || {
    for item in myvec.iter() {
        if something(item) { continue; }
        if let Some(foo) = something_else(item) { yield foo; }
    }
});
```

and that will require holding a borrow across a yield.

nikomatsakis: > The pinning also applies to the design of AsyncRead/AsyncWrite, which currently use Pin even though there's no clear plan to make them implemented with generator type syntax. In essence the asyncification of a singature is currently understood as pinned reciever + context arg + return poll.

Yes. I'm trying to push a bit on this and make sure I understand *why* we think that is the correct approach, I guess. =)

nikomatsakis: So: I don't think that this conversation has to block this PR from landing, and it probably shouldn't even block the RFC itself, but I stll feel like there is *some* missing text around the `next` method that at minimum explains *why* `Unpin` is needed and perhaps tries to cover some of the reasoning above about why `poll_next` should take a pinned argument. Maybe the right place for some of this text is in the coverage of generator syntax, since we can discuss some of the difficulties that will have to be resolved in making generator syntax (and async generator syntax) work.

withoutboats: > With respect to generators that cannot borrow across a yield, I think it's not going to work. I imagine a lot of people are going to want to write code like this

See unless you write `gen move`, this is not a self-referential type at all, because it can capture `myvec` by reference. :) Users writing futures combinators *immediately* ran into the problem of them being best expressed by self-referential types (and ended up having to Arc<Mutex> all kinds of stuff to share ownership across yield points), but users writing iterator combintors never have that problem. I think this is worth considering more seriously.

withoutboats: The key to the difference is that futures are ultimately passed to some executor API like `spawn`, which expects a static future; to achieve that the futures contain all the state they need, and references are internal to that state. Iterators are almost never required to be `'static` by the APIs that consume them.

However, it's still the case that it will be far more natural to end up in self-referential situations with a generator than it is with a manually implemented iterator, because you will be able to implicitly build that state using the member variables of a function. So the restriction might still limit users from creating code in a natural way. And it's a bit confusing for the restriction to exist only in non-async generators.

So we'll have to get more experience to figure it out.

nikomatsakis:
I'm going to go ahead and merge these edits, since I think they've been at a stable point for some time.

PR #17, opened 2020-08-04
Update stream rfc

nellshamrell:
[Rendered](https://github.com/rust-lang/wg-async-foundations/blob/158cce517090a5c7648d1ca7576f3c6d0741ca22/rfc-drafts/stream.md)

### DIFF
--- a/rfc-drafts/stream.md
+++ b/rfc-drafts/stream.md
@@ -94,7 +94,7 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
 
 When drafting this RFC, there was a [good deal of discussion](https://github.com/rust-lang/wg-async-foundations/pull/15#discussion_r452482084) around why the `next` method requires `Self:Unpin`.
 
-To understand this, it helps to take a closer look at the definition of `Next` (this struct is further discussed later in this RFC) in the [futures-util crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12)).
+To understand this, it helps to take a closer look at the definition of `Next` (this struct is further discussed later in this RFC) in the [futures-util crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12).
 
 ```rust
 pub struct Next<'a, St: ?Sized> {
@@ -235,7 +235,7 @@ Stream` values without the need to monomorphize the functions that work
 with them.
 
 Unfortunately, the use of poll does mean that it is harder to write
-stream implementations. The long-term fix for this, discussed in the [Future possibilities][future-possibilities] section, is dedicated [generator syntax].
+stream implementations. The long-term fix for this, discussed in the [Future possibilities] section, is dedicated [generator syntax].
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -684,7 +684,7 @@ gen_fn().pin_somehow().adapter1().adapter2()
 
 With streams, the core interface _is_ pinned, so pinning occurs at the last moment.
 
-The general shap would be 
+The general shape would be 
 
 ```rust
 async_gen_fn().adapter1().adapter2().pin_somehow()
--- a/rfc-drafts/stream.md
+++ b/rfc-drafts/stream.md
@@ -106,7 +106,7 @@ Since `Stream::poll_next` takes a pinned reference, the next future needs `S` to
 
 An alternative approach we could take would be to have the `next` method take `Pin<&mut S>`, rather than `&mut S`. However, this would require pinning even when the type is `Unpin`. The current approach requires pinning only when the type is not `Unpin`.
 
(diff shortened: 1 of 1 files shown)


rfc-drafts/stream.md
```diff
@@ -93,6 +90,24 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
...
+
+```rust
+pub struct Next<'a, St: ?Sized> {
+    stream: &'a mut St,
+}
+```
+
+Since `Stream::poll_next` takes a pinned reference, the next future needs `S` to be `Unpin` in order to safely construct a `Pin<&mut S>` from a `&mut S`.
+
+An alternative approach we could take would be to have the `next` method take `Pin<&mut S>`, rather than `&mut S`. However, this would require pinning even when the type is `Unpin`. The current approach requires pinning only when the type is not `Unpin`.
```
Matthias247, line 107: I'm wondering what the real practical implications are.

At the end users want to write `stream.next().await` - and not using the manual poll methods. In order to make this viable the Stream must be `Unpin` with this approach. Most streams are that in the current world, others can be stack-pinned. If we had `next(Pin<&mut Self>)`, the practical implication is that this pinning is mandatory, as outlined here. Certainly a big usability hit if every use would end up with:

```
async fn do_with_stream(stream: Stream<Something>) {
    pin_mut!(stream);
    let item = stream.next().await();
}
``` 

This also then also prevents passing Streams around between usages, unless only the `Pin`s are passed (which ends up being an ergonomic nightmare due to pin projections).

**=> Overall I agree, I don't see any benefit of letting `Pin<&mut Self>` in favor of `&mut self` and `S: Unpin`**

For the question whether `poll_next()` on `Stream` should simple take `&mut self` I'm still partial. For hand written Streams, which are currently the majority, `&mut self` will always be a lot more ergonomic. And if we go for that the question around requiring `Unpin` for `StreamExt::next` never comes up. Streams which abstract over `async fn`s would require boxing then however. For desktop applications I don't see that as an acceptable tradeoff. For embedded it might be a bit annoying, but the big `Future` and `Stream` sizes are a problem there anyway. For someone wanting to write hot loops with `Stream`s it's also a problem - but you can always fall back to just using `async fn`s  for that.

rfc-drafts/stream.md
```diff
@@ -634,4 +649,66 @@ If we introduce `-> Stream` first, we will have to permit `LendingStream` in the
...
+
+With streams, the core interface _is_ pinned, so pinning occurs at the last moment.
+
+The general shape would be 
+
+```rust
+async_gen_fn().adapter1().adapter2().pin_somehow()
+```
+
+Pinning at the end, like with a stream, lets you build and return those adapters and then apply pinning at the end. This may be the more efficient setup and implies that, in order to have a `gen fn` that produces iterators, we will need to potentially disalllow borrowing yields or implement some kind of `PinnedIterator` trait that can be "adapted" into an iterator by pinning.
```
Matthias247, line 693: Typo fix

```suggestion
Pinning at the end, like with a stream, lets you build and return those adapters and then apply pinning at the end. This may be the more efficient setup and implies that, in order to have a `gen fn` that produces iterators, we will need to potentially disallow borrowing yields or implement some kind of `PinnedIterator` trait that can be "adapted" into an iterator by pinning.
```

rfc-drafts/stream.md
```diff
@@ -634,4 +649,66 @@ If we introduce `-> Stream` first, we will have to permit `LendingStream` in the
 Additionally, if we introduce `LendingStream` later, we'll have to figure out how
 to convert a `LendingStream` into a `Stream` seamlessly.
 
-Further designing generator functions is out of the scope of this RFC.
+### Differences between Iterator generators and Async generators
```
Matthias247, line 652: I feel like this whole section helps more the language theorists than the end users among us.

I am pretty much one of the experts in my company for async programming (in Rust and other languages), but have a hard time on figuring out what it wants to convey to me. That there is a difference in pinning is understandable (even though I now 95% of users won't know what pinning really is). But what would I use `PinIterator` for? And why? Maybe leave it out.

I would also encourage again to think about `Stream`s not as a tool for crazy combinators.
The goal of async/await is explicitely to allow people to just write "plain old control flow". If you need combinators - write a for loop. Everyone can do that, and it's in most cases a lot easier for reviewers to figure out whats going on in them than with combinators. If we have any limitations on combinators I still don't think it's a huge loss for practial purposes.

nellshamrell: There will be plenty of language theorists commenting on the RFC - this is me trying to get ahead of that. And combinators will also most definitely come up in the comments (it certainly has in our pre-RFC discussions). One of the goals of the RFC process is to show that we have throught through as many of the implications as we possibly can. Even though it's not relevant to you personally (or even most of the audience) I would like to leave it in for the few (but vocal!) for whom it is relevant.

nellshamrell:
Closing in favor of #18

PR #18, opened 2020-09-14
adds updates based on lang team meeting 2020-08-26

nellshamrell:
[Rendered](https://github.com/rust-lang/wg-async-foundations/blob/aa01041d7ee96b2306108c9af8f29293b0cfe174/rfc-drafts/stream.md)

[Language team notes](https://github.com/rust-lang/lang-team/blob/master/design-meeting-minutes/2020-08-26-stream-trait.md) on async stream design.

Signed-off-by: Nell Shamrell <nellshamrell@gmail.com>

### DIFF
--- a/rfc-drafts/stream.md
+++ b/rfc-drafts/stream.md
@@ -73,6 +73,7 @@ pub trait Stream {
     fn next(&mut self) -> Next<'_, Self>
     where
         Self: Unpin;
+    { ... }
 }
 ```
 
--- a/rfc-drafts/stream.md
+++ b/rfc-drafts/stream.md
@@ -106,7 +106,7 @@ Since `Stream::poll_next` takes a pinned reference, the next future needs `S` to
 
 An alternative approach we could take would be to have the `next` method take `Pin<&mut S>`, rather than `&mut S`. However, this would require pinning even when the type is `Unpin`. The current approach requires pinning only when the type is not `Unpin`.
 
-At the moment, we do not see many `Unpin!` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [future-possibilities].
+At the moment, we do not see many `!Unpin` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [future-possibilities].
 
 In summary, an async stream:
 * has a pinned receiver
--- a/rfc-drafts/stream.md
+++ b/rfc-drafts/stream.md
@@ -91,28 +91,6 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
 [context]: https://doc.rust-lang.org/std/task/struct.Context.html
 [Waker]: https://doc.rust-lang.org/std/task/struct.Waker.html
 
-### Why does next require Self:Unpin?
-
-When drafting this RFC, there was a [good deal of discussion](https://github.com/rust-lang/wg-async-foundations/pull/15#discussion_r452482084) around why the `next` method requires `Self:Unpin`.
-
-To understand this, it helps to take a closer look at the definition of `Next` (this struct is further discussed later in this RFC) in the [futures-util crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12).
-
-```rust
-pub struct Next<'a, St: ?Sized> {
-    stream: &'a mut St,
-}
-```
-Since `Stream::poll_next` takes a pinned reference, the next future needs `S` to be `Unpin` in order to safely construct a `Pin<&mut S>` from a `&mut S`.
-
-An alternative approach we could take would be to have the `next` method take `Pin<&mut S>`, rather than `&mut S`. However, this would require pinning even when the type is `Unpin`. The current approach requires pinning only when the type is not `Unpin`.
-
-At the moment, we do not see many `!Unpin` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [Future possibilities](future-possibilities).
-
-In summary, an async stream:
-* has a pinned receiver
-* that takes `cx` context
-* and returns a `Poll`
-
 ## Initial impls
 
 There are a number of simple "bridge" impls that are also provided:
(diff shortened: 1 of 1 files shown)


rfc-drafts/stream.md
```diff
@@ -93,6 +91,28 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
...
+```rust
+pub struct Next<'a, St: ?Sized> {
+    stream: &'a mut St,
+}
+```
+Since `Stream::poll_next` takes a pinned reference, the next future needs `S` to be `Unpin` in order to safely construct a `Pin<&mut S>` from a `&mut S`.
+
+An alternative approach we could take would be to have the `next` method take `Pin<&mut S>`, rather than `&mut S`. However, this would require pinning even when the type is `Unpin`. The current approach requires pinning only when the type is not `Unpin`.
+
+At the moment, we do not see many `Unpin!` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [future-possibilities].
```
programmerjake, line 109:
```suggestion
At the moment, we do not see many `!Unpin` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [future-possibilities].
```

rfc-drafts/stream.md
```diff
@@ -183,6 +203,21 @@ while let Some(v) = stream.next().await {
 }
 ```
 
+We could also consdier adding a try_next? function, allowing
```
programmerjake, line 206:
```suggestion
We could also consider adding a try_next? function, allowing
```

nellshamrell:
Good catches, @programmerjake, I committed those changes :)

rfc-drafts/stream.md
```diff
@@ -93,6 +91,28 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
...
+To understand this, it helps to take a closer look at the definition of `Next` (this struct is further discussed later in this RFC) in the [futures-util crate](https://docs.rs/futures-util/0.3.5/src/futures_util/stream/stream/next.rs.html#10-12).
+
+```rust
+pub struct Next<'a, St: ?Sized> {
+    stream: &'a mut St,
+}
+```
+Since `Stream::poll_next` takes a pinned reference, the next future needs `S` to be `Unpin` in order to safely construct a `Pin<&mut S>` from a `&mut S`.
+
+An alternative approach we could take would be to have the `next` method take `Pin<&mut S>`, rather than `&mut S`. However, this would require pinning even when the type is `Unpin`. The current approach requires pinning only when the type is not `Unpin`.
```
tmandry, line 107: I wonder if it's worth reminding people that if you already have a `Pin<&mut S>` you can access `.next()` (through the bridge impl below) because `Pin<P>: Unpin`.

rfc-drafts/stream.md
```diff
@@ -93,6 +91,28 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
...
+
+At the moment, we do not see many `!Unpin` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [future-possibilities].
+
+In summary, an async stream:
+* has a pinned receiver
+* that takes `cx` context
+* and returns a `Poll`
```
tmandry, line 114: This is a summary of `poll_next()` but not the whole section, which I find a tad confusing.

rfc-drafts/stream.md
```diff
@@ -93,6 +91,28 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
...
+```rust
+pub struct Next<'a, St: ?Sized> {
+    stream: &'a mut St,
+}
+```
+Since `Stream::poll_next` takes a pinned reference, the next future needs `S` to be `Unpin` in order to safely construct a `Pin<&mut S>` from a `&mut S`.
+
+An alternative approach we could take would be to have the `next` method take `Pin<&mut S>`, rather than `&mut S`. However, this would require pinning even when the type is `Unpin`. The current approach requires pinning only when the type is not `Unpin`.
+
+At the moment, we do not see many `!Unpin` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [future-possibilities].
```
tmandry, line 109:
```suggestion
At the moment, we do not see many `!Unpin` streams in practice (though there is one in the [futures-intrusive crate](https://github.com/Matthias247/futures-intrusive/blob/master/src/channel/mpmc.rs#L565-L625)). Where they will become important is when we introduce async generators, as discussed in [Future possibilities](future-possibilities).
```

rfc-drafts/stream.md
```diff
@@ -93,6 +91,28 @@ The arguments to `poll_next` match that of the [`Future::poll`] method:
 [context]: https://doc.rust-lang.org/std/task/struct.Context.html
 [Waker]: https://doc.rust-lang.org/std/task/struct.Waker.html
 
+### Why does next require Self:Unpin?
```
tmandry, line 94: I'm not sure this belongs in the Guide-level explanation, it seems like it'd be a better fit in the section dedicated to Next below.

rfc-drafts/stream.md
```diff
@@ -183,6 +203,21 @@ while let Some(v) = stream.next().await {
 }
 ```
 
+We could also consider adding a try_next? function, allowing
```
tmandry, line 206: I had to [look this up](https://docs.rs/futures/0.3.5/futures/stream/trait.TryStreamExt.html#method.try_next) to remember why it was useful, so it might be worth linking to the futures-rs version.

tmandry: I'd also expect to see some reasoning on why this isn't being proposed by the RFC yet. We want to stick to just the bits everyone needs for now, or something along those lines.

rfc-drafts/stream.md
```diff
@@ -483,6 +515,16 @@ We may wish to extend the `for` loop so that it works over streams as well.
 for elem in stream { ... }
 ```
 
+One of the complications of using `while let` syntax is the need to pin.
+A `for` loop syntax that takes ownership of the stream would be able to
+do the pinning for you. 
+
+We may not want to make sequential processing "too easy" without also enabling
```
tmandry, line 522: (suggestion) On the other hand,

rfc-drafts/stream.md
```diff
@@ -634,4 +684,75 @@ If we introduce `-> Stream` first, we will have to permit `LendingStream` in the
...
+}
+impl<I: PinIterator, P: Deref<Target = I> + DerefMut> Iterator for Pin<P> {
+    type Item = <I as PinIterator>::Item;
+    fn next(&mut self) -> Self::Item { self.as_mut().next() }
+}
+
+// this would be nice.. but would lead to name resolution ambiguity for our combinators 😬 
+default impl<T: Iterator> PinIterator for T { .. }
+```
+Pinning also applies to the design of AsyncRead/AsyncWrite, which currently uses Pin even through there is no clear plan to make them implemented with generator type syntax. The asyncification of a signature is current understood as pinned receiver + context arg + return poll.
```
tmandry, line 746:
```suggestion
Pinning also applies to the design of AsyncRead/AsyncWrite, which currently uses Pin even through there is no clear plan to make them implemented with generator type syntax. The asyncification of a signature is currently understood as pinned receiver + context arg + return poll.
```

PR #29, opened 2021-03-05
Fix some broken links.

ehuss:
I noticed a few links that didn't work, I think these are correct.

Also, adding site-url allows mdbook's 404 support to work correctly when deployed on the website.


tmandry:
Thanks!

PR #30, opened 2021-03-05
integrate new character axes and applications

nikomatsakis:
Update content somewhat, integrating the idea of character axes and applications.

Rendered form: https://nikomatsakis.github.io/wg-async-foundations/


tmandry:
I don't think you meant to commit all these rendered files?

tmandry:
I think we should keep these as short and to the point as possible. Here's a draft for operating systems:

> Core system services like filesystems and networking. IPC between various programs and services. Kernel and driver code; waiting on device I/O.
> - What separates this from others?
>     - Latency profile: Typically 100's-1,000's of ns, uncontended, for IPC

Since the list is kind of long maybe we should only keep skeletons around for the ones we _know_ we want and include a list of bullet points at the end for skimming.

I think database servers and high-performance I/O are redundant.

nikomatsakis:
@tmandry I did not mean to commit rendered files, it's true :)

Also, I agree the list is long. I like the idea of having a few with short descriptions, and a bullet, and encouraging folks to turn them into something more real.

nikomatsakis:
I've reworked the presentation of the axes to emphasize the characters more. I think it was kind of overwhelming before.

nikomatsakis:
Gonna merge this so we can keep iterating =)

PR #33, opened 2021-03-10
Expand Grace and networking

carllerche:
Talks a bit about coming to Rust w/ existing networking experience.

### DIFF
--- a/src/vision/status_quo.md
+++ b/src/vision/status_quo.md
@@ -123,22 +123,31 @@ Outline:
 
 [Grace deploys her service and is able to fix problems]: ./shiny_future.md#grace-deploys-her-service-and-is-able-to-fix-problems
 
-When examining metrics for her service in the past, Grace has noticed tail latencies in the P99 that exceed their SLA, and identified GC in the routing layer as the problem. She decides to investigate rewriting the routing service in Rust. A few months of frantic hacking follow:
+When examining her service metrics, Grace notices tail latencies in the P99 that exceed their target. She identifies GC in the routing layer as the culprit. Grace follows industry trends and is already aware of Rust and its ecosystem at a high level. She decides to investigate rewriting the routing service in Rust.
+
+To meet throughput requirements, Grace has already decided to use a thread-per-core model and minimize cross-thread communication. She explores available ecosystem options and finds no option that gets her exactly what she is looking for out of the box. However, she can use Tokio with minimal configuration to achieve her architecture.
+
+A few months of frantic hacking follow.
 
 <img src="https://media.giphy.com/media/ule4vhcY1xEKQ/giphy.gif" alt="montage of cats typing" width=200></img>
  
-Soon enough, she and her team have a proof of concept working. They run some local stress tests and notice that 5% of requests hang and fail to respond and the client eventually times out. She is unable to reproduce this problem when running one-off requests locally, it only shows up when sending above 200 requests-per-second. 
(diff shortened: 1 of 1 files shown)


src/vision/status_quo.md
```diff
@@ -123,22 +123,31 @@ Outline:
 
 [Grace deploys her service and is able to fix problems]: ./shiny_future.md#grace-deploys-her-service-and-is-able-to-fix-problems
 
-When examining metrics for her service in the past, Grace has noticed tail latencies in the P99 that exceed their SLA, and identified GC in the routing layer as the problem. She decides to investigate rewriting the routing service in Rust. A few months of frantic hacking follow:
+When examining her service metrics, Grace notices tail latencies in the P99 that exceed their target. She identifies GC in the routing layer as the culprit. Grace follows industry trends and is already aware of Rust and its ecosystem at a high level. She decides to investigate rewriting the routing service in Rust.
+
+To meet throughput requirements, Grace has already decided to use a thread-per-core model and minimize cross-thread communication. She explores available ecosystem options and finds no option that gets her exactly what she is looking for out of the box. However, she can use Tokio with minimal configuration to achieve her architecture.
```
nikomatsakis, line 128: I feel like this section could be expanded and turned into a story of its own. Like, how did she do that research? Was it easy to find out what her options are? Easy to find out how configurable each runtime is?

I'm torn about citing tokio here. On the one hand, I want these stories to be real, and this one did come from a tokio user. On the other hand, it's not particularly important to the story itself. One question that came up for me was whether other runtimes would be able to do similar configuration (perhaps Grace didn't know how to do it) -- that's something I would at minimum want to see in a FAQ, but I'm thinking maybe it's easier to just elide the details and say "Eventually she settles on a runtime that she can configure to meet her needs."

nikomatsakis (approved):
Thanks Carl! If it's ok with you, I'd like to make a few edits then merge this, then perhaps split it into two status quo stories.

src/vision/status_quo.md
nikomatsakis: After some reflection, I like the idea of being specific, but I think we should add a FAQ like "Could Grace have used another runtime to achieve the same objectives?" with an answer like, "Maybe! But in this instance the people this story is based on were using tokio, so that's the one we wrote into the story." (Folks with more specific knowledge could expand that answer at some point.)

nikomatsakis:
I'm gonna merge and make those edits.

src/vision/status_quo.md
```diff
@@ -123,22 +123,31 @@ Outline:
...
 
 * **What are the morals of the story?**
-    * (Explain your key points here)
-* **Are you just making this stuff up?**
-    * (Cite important sources here; feel free to add files into the conversations folder for more details)
+    * When coming from a background of network engineering, users will bring their own design choices around architecture.
+        * Examples: [seastar](http://seastar.io/) and [Glommio](https://www.datadoghq.com/blog/engineering/introducing-glommio/)
+    * There is a lack of debugging tools for async.
+    * Writing futures by hand is error prone.
+* **What are the sources for this story??**
```
bl-ue, line 149: Two question marks here—is this intended? Added by 9e18afe0bd72a90f712bf46e946607d9631b5975.

nikomatsakis: Not intended :)

src/vision/status_quo.md
carllerche: I'm not aware of any other option available today.

PR #41, opened 2021-03-17
create example projects, independent from characters

nikomatsakis:
As discussed on Zulip

[Rendered preview.](https://nikomatsakis.github.io/wg-async-foundations/)


nikomatsakis:
Alternative to #39

PR #48, opened 2021-03-18
status quo: Alan hates writing a Stream

seanmonstar:
[Rendered](https://github.com/seanmonstar/wg-async-foundations/blob/status-quo-streaming-yuck/src/vision/status_quo/alan_hates_writing_a_stream.md)

### DIFF
--- a/src/vision/status_quo/alan_hates_writing_a_stream.md
+++ b/src/vision/status_quo/alan_hates_writing_a_stream.md
@@ -12,7 +12,7 @@ If you would like to expand on this story, or adjust the answers to the FAQ, fee
 
 After a couple weeks learning Rust basics, Alan quickly understands `async` and `await`, and therefore has several routes built for his application that await a few things and then construct an HTTP response and send a buffered body. To build the buffered response bodies, Alan was reading a file, and then appending a signature, and putting that all into a single buffer of bytes.
 
-Eventually, Alan realizes that some responses have enormous bodies, and would like stream them instead of buffering them fully in memory. He's *used* the `Stream` trait before. Using it was very natural, and followed a similar pattern to regular `async`/`await`:
+Eventually, Alan realizes that some responses have enormous bodies, and would like to stream them instead of buffering them fully in memory. He's *used* the `Stream` trait before. Using it was very natural, and followed a similar pattern to regular `async`/`await`:
 
 ```rust
 while let Some(chunk) = body.next().await? {
@@ -61,11 +61,11 @@ error[E0658]: yield syntax is experimental
    = note: see issue #43122 <https://github.com/rust-lang/rust/issues/43122> for more information
 ```
 
-After reading about how yield is experimental, and giving up reading the 100+ comments in the linked issue, Alan figures he's just got to implement `Stream` manually.
+After reading about how yield is experimental, and giving up reading the 100+ comments in the [linked issue](https://github.com/rust-lang/rust/issues/43122), Alan figures he's just got to implement `Stream` manually.
 
 ### Implementing `Stream`
 
-Implementing a `Stream` means writing async code in a way that doesn't _feel_ like all the other `async` code that the user has already written. The user is introduced to writing a `poll` function, which forces these three things into their face:
+Implementing a `Stream` means writing async code in a way that doesn't _feel_ like the `async fn` that Alan has written so far. He needs to write a `poll` function and it has a lot of unfamiliar concepts:
 
 - `Pin`
 - State machines
@@ -89,7 +89,7 @@ impl Stream for SigningFile {
 
 #### Pin :scream:
 
-First, he notices `Pin`. Alan wonders, "Why does `self` have bounds? I've only ever seen `self`, `&self`, and `&mut self` before". Curious, he reads the `std::pin` page, and a bunch of jargon about pinning data in memory. He also reads that this is useful to guarantee that an object cannot move, and he wonders why he cares about that. The only example on the page explains how to write a ["self-referential struct"][self-ref], but notices it needs `unsafe` code, and that triggers an internal alarm in Alan: "I thought Rust was safe..."
+First, he notices `Pin`. Alan wonders, "Why does `self` have bounds? I've only ever seen `self`, `&self`, and `&mut self` before". Curious, he reads the [`std::pin`](https://doc.rust-lang.org/std/pin/struct.Pin.html) page, and a bunch of jargon about pinning data in memory. He also reads that this is useful to guarantee that an object cannot move, and he wonders why he cares about that. The only example on the page explains how to write a ["self-referential struct"][self-ref], but notices it needs `unsafe` code, and that triggers an internal alarm in Alan: "I thought Rust was safe..."
 
 Eventually (how?), Alan realizes that the types he's depending on are `Unpin`, and so he doesn't need to worry about the unsafe stuff. It's just a more-annoying pointer type.
 
@@ -174,7 +174,7 @@ A little later, Alan needs to add some response body transforming to some routes
 
 * **What are the morals of the story?**
     * Writing an async `Stream` is drastically different than writing an `async fn`.
-    * `Pin` is explained in an abstract way, but do most users who run into `Pin` for the first time need all that? It looks so scary.
+    * The documentation for `Pin` doesn't provide much practical guidance in how to use it, instead focusing on more abstract considerations.
     * Missing a waker registration is a runtime error, and very hard to debug. If it's even possible, a compiler warning or hint would go a long way.
 * **What are the sources for this story?**
     * Part of this story is based on the original motivation for `async`/`await` in Rust, since similar problems exist writing `impl Future`.
--- a/src/vision/status_quo/alan_hates_writing_a_stream.md
+++ b/src/vision/status_quo/alan_hates_writing_a_stream.md
@@ -1,0 +1,194 @@
+# 😱 Status quo stories: Alan hates writing a `Stream`
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+[Alan][] is used to writing web server applications using async sockets, but wants to try Rust to get that signature vroom vroom.
+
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+A little later, Alan needs to add some response body transforming to some routes, to add some app-specific framing. Upon realizing he needs to implement another `Stream` in a generic fashion, he instead closes the editor and complains on Twitter.
+
+
+
+## 🤔 Frequently Asked Questions
+
+
+* **What are the morals of the story?**
+    * Writing an async `Stream` is drastically different than writing an `async fn`.
+    * `Pin` is explained in an abstract way, but do most users who run into `Pin` for the first time need all that? It looks so scary.
```
nikomatsakis, line 177:
```suggestion
    * The documentation for `Pin` doesn't provide much practical guidance in how to use it, instead focusing on more abstract considerations.
```

nikomatsakis (approved):
Thanks @seanmonstar-- I left a few suggestions and thoughts.

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+
+Eventually, Alan realizes that some responses have enormous bodies, and would like to stream them instead of buffering them fully in memory. He's *used* the `Stream` trait before. Using it was very natural, and followed a similar pattern to regular `async`/`await`:
+
+```rust
+while let Some(chunk) = body.next().await? {
+    file.write_all(&chunk).await?;
+}
+```
+
+However, _implementing_ `Stream` turns out to be rather different. While he quickly learned the simple way to turn a `File` into a `Stream` with `StreamReader`, the chaining part was much harder.
```
rylev, line 23: Question: how did Alan learn about `StreamReader`? Were the docs for chaining streams not mentioned in those docs? It would be interesting to figure out where the breakdown in docs was exactly.

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+
+Eventually (how?), Alan realizes that the types he's depending on are `Unpin`, and so he doesn't need to worry about the unsafe stuff. It's just a more-annoying pointer type.
+
+[self-ref]: https://doc.rust-lang.org/std/pin/index.html#example-self-referential-struct
+
+#### State Machine
+
+With `Pin` hopefully ignored, Alan next notices that in the imperative style he wanted originally, he didn't need to explicitly keep track of state. The state was simply the imperative order of the function. But in a `poll` function, the state isn't saved by the compiler. Alan finds blog posts about the dark ages of Futures 0.1, when even `Future` was written with a "state machine".
```
rylev, line 100: nit: I had to re-read the last sentence twice to get what you mean. It's not that the `Future` trait itself was defined with a state machine, but rather that futures (i.e., types that implemented `Future`) were written using state machines (and in fact still are if you manually implement `Future` for a type). Perhaps this could be worded slightly differently to avoid this hiccup in understanding?

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+}
+```
+
+Now he tries to write the `poll_next` method, checking readiness of individual steps and proceeding to the next state, while grumbling away the weird `Pin` noise:
+
+```rust
+todo!("is this worth showing in the story, or just noise?")
+```
+
+It turns out it was more complicated than Alan thought (the author made this same mistake). The `digest` method of `Signature` is `async`, _and_ it consumes the signature, so the state machine needs to be adjusted. The signature needs to be able to be moved out, and it needs to be able to store a future from an `async fn`. Trying to figure out how to represent that in the type system was difficult (expand?), but Alan eventually learns to just store a `Pin<Box<dyn Future>>`, wondering if the `Pin` there is important.
```
rylev, line 123: At the risk of the story getting too long, I do think it's worth expanding how Alan was able to land on `Pin<Box<dyn Future>>`. `Pin<Box<dyn Future>>`  involves pinning and dynamic dispatch (which we need presumably since the future from `digest` can't be named. This is confusing stuff for relative newcomers!

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+}
+
+enum State {
+    File(ReaderStream, Signature),
+    Signing(Pin<Box<dyn Future<Item = Vec<u8>>>>),
+    Done,
+}
+```
+
+(show how gross this codes gets?)
```
rylev, line 137: Yes!

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+    },
+    None => {
+        me.state = State::Sign;
+        // oops!
+        return Poll::Pending;
+    }
+}
+```
+
+In one of the branches, the state is changed, and `Poll::Pending` is returned. Alan assumes that the task will be polled again with the new state. But, since the file was done (and has returned `Poll::Ready`), there was actually no waker registered to wake the task again. So his stream just hangs forever.
```
rylev, line 160: The code uses the `ready!` macro. How did Alan land on using that? Did its use potentially obscure things even more?

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+The compiler doesn't help at all, and he re-reads his code multiple times, but because of this easy-to-misunderstand logic error, Alan eventually has to ask for help in a chat room. After a half hour of explaining all sorts of details, a kind person points out he either needs to register a waker, or perhaps use a loop.
+
+(Show the solution? in far too many cases (likely this one), the code gets turned into a weird loop, or the user must learn about wakers).
+
+### Gives Up
+
+A little later, Alan needs to add some response body transforming to some routes, to add some app-specific framing. Upon realizing he needs to implement another `Stream` in a generic fashion, he instead closes the editor and complains on Twitter.
+
+
+
```
rylev, line 171: Nit: can you remove some this space?

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+## 🤔 Frequently Asked Questions
+
+
+* **What are the morals of the story?**
+    * Writing an async `Stream` is drastically different than writing an `async fn`.
```
rylev, line 176: I'm worried that this story could be shrugged off as "landing generators will fix this". Can we also make a note that the primitives that Rust's async story uses are not super straight-forward to compose?

rylev (commented):
Awesome stuff! 🎉 I think the areas where you were unsure whether they should be expanded, should be expanded. This story is very illustrative.

src/vision/status_quo/alan_hates_writing_a_stream.md
rylev: Also, don't you mean `ReaderStream`?

src/vision/status_quo/alan_hates_writing_a_stream.md
seanmonstar: It's probable that generators are indeed the "shiny future", but there's two more bullet points about `Pin` and `Waker`. What else about "primitives" did you have in mind as morals?

nikomatsakis: I was thinking the same thing, I expect generators *are* a big part of the shiny future. I guess that one thing which might help would be to write a second story, or expand this story, to talk about implementation a custom future -- or perhaps add a FAQ like "Besides streams, where else might users encounter pin, context, and friends?"

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,194 @@
...
+   |         ^^^^^^^^^^^^^^^
+   |
+   = note: see issue #43122 <https://github.com/rust-lang/rust/issues/43122> for more information
+```
+
+After reading about how yield is experimental, and giving up reading the 100+ comments in the [linked issue](https://github.com/rust-lang/rust/issues/43122), Alan figures he's just got to implement `Stream` manually.
+
+### Implementing `Stream`
+
+Implementing a `Stream` means writing async code in a way that doesn't _feel_ like the `async fn` that Alan has written so far. He needs to write a `poll` function and it has a lot of unfamiliar concepts:
```
sehz, line 68: In my opinion, `Stream` trait is very low level and we should not encourage most Rust developer to use it directly.  Instead, we should teach them to use combinators and other high level primitives (map, filter, etc)  to construct Stream.  

Teaching `poll_next` should be reserved for advance use cases (i.g. writing device drivers, protocol encoders)

src/vision/status_quo/alan_hates_writing_a_stream.md
seanmonstar: I tried expanding a little, about trying out a generic and realizing that wouldn't work, and a `Signing(impl Future)` isn't stable yet, and then finally asking Barbara for help again. 🤷

src/vision/status_quo/alan_hates_writing_a_stream.md
seanmonstar: I've updated that Alan remembers `ready!` from the futures 0.1 blog posts. I don't believe it obscures things (but I could be blind by my own knowledge...)

seanmonstar:
Updated with more code and made some sentences clearer. 

(I was pleasantly surprised that the code was marginally less gross, since now the compiler lets me move even inside a match arm like `State::File(ref mut file) => { .. }`, since it can see I don't touch `file` again. That didn't used to be the case :D)

src/vision/status_quo/alan_hates_writing_a_stream.md
```diff
@@ -0,0 +1,244 @@
...
+
+Oh well, at least it _works_, right?
+
+### Wakers
+
+So far, Alan hasn't paid too much attention to `Context` and `Poll`. It's been fine to simply pass them along untouched. There's a confusing bug in his state machine. Let's look more closely:
+
+```rust
+// zooming in!
+match ready!(file.poll_next(cx)) {
```
nikomatsakis, line 171:
```suggestion
match ready!(Pin::new(file).poll_next(cx)) {
```

nikomatsakis (changes requested):
This is awesome @seanmonstar! I left one nit but I'd be up to merge.

rylev (approved):
Looks great!

src/vision/status_quo/alan_hates_writing_a_stream.md
Matthias247: I would rather think about a shiny future which doesn't involve more special types but less. The whole thing could also be avoided by 

```rust
trait Stream {
    type Item;
    async fn next(&mut self) -> Option<Self::Item>;
}
```

We are all aware that this one also has challenges. But it applies for all composed async operations.

But since the goal of this doc is to showcase the challenge and not solutions, it seems good as is.

PR #66, opened 2021-03-19
status quo: GUI Framework

Fice:
Hi, I'm currently experimenting with writing a GUI Abstraction framework (meaning I want the OS to render buttons and stuff) and
I made some nice progress with how I want the API to look and feel and how to make it work in rust. Yesterday I finally started on how to deal with user events (e.g. mouse clicks).
My background is actually C/C++ but I recently did a few Typescript/AngularJS Frontend Projects. I really liked how Angular uses [Observables](https://rxjs.dev/guide/observable) for dealing with events and that's why I would really like to do something similar in Rust. I'm currently looking into and playing around with Streams and it looks promising.

I tried to keep the project described in the PR a generic GUI frameworks with the only assumption that it uses Streams for the user event handling.

I did add some example code on how such an API could be used. It's meant to give an idea and is not meant to compile. I hope this makes things a bit more clear.

## Character
I do believe a new character might be useful. Someone with a background in Javascript or similar with a focus on **Frontend-Applications**. I feel the current character are more Backend/Low-Level and not Front-End Developers. I know this is not a common scenario with Rust missing a UI Framework but that hopefully changes in the future. Let me know and I can write something up.

## Note:
I probably mixed up some stuff. Some parts are written from the view of someone *Developing a UI framework* in rust while other aspects come from someone *using a UI framework* in rust.
I'm not sure if it's worth splitting those two up.


nikomatsakis:
@Fice Alan has a background in JavaScript, I imagine he would be a good fit for your purposes.

src/vision/projects/StrUI
```diff
@@ -0,0 +1,48 @@
+# ⚡ Projects: Strui (GUI Framework)
+
+## What is this?
+This is a sample project for use within the various "status quo" or "shiny future" stories.
+
+## Description
+Finally a Rust GUI Framework for the masses :)
+
+It uses Streams as a way of Event Handling:
```
nikomatsakis, line 9: Hmm, I'm considering. This feels really different from the other projects in that it is kind of half-way to a user story, and has code examples and the like. I would be curious to get an idea what stories you'd like to write that are tied to this project. 

I do think having a project representing GUIs or other use cases is a good idea.

Fice: Not sure why it didn't cross my mind, but it makes absolute sense to use a GUI Application as a Project. It's more generic and I'm sure I can bring across the points I wanted to make. Will rewrite tomorrow.

I have quite a few *shiny future* scenarios in mind.

The only current one is:
I'm talking about this Trait, not sure if this is in scope: https://docs.rs/futures/0.3.13/futures/stream/trait.StreamExt.html
If you have some sort of middleware, that works by getting a stream, manipulating it (using ```filter```, ```map``` or similar) and returning that manipulated stream. All well and good, except when you shorten it's lifetime like ```some_stream.take(1)```. All those stream manipulations that return a stream, that indicates it's done sooner than the original one, are most likely the wrong choice in my scenario and can wreck the whole system. Would be nice to make sure a user cannot accidentally use those.

Fice:
Alan looked like a Backend Developer to me.
> Ruby on Rails, node.js, and used Django

Altough Frontend stuff is quite common for node.js guys so I'm happy with him if you are :)

nikomatsakis:
@Fice When it comes to JS or what have you, I guess I was imagining more "full stack". Maybe we can tweak the bio to make that clear. :)

nikomatsakis:
@Fice 

> Not sure why it didn't cross my mind, but it makes absolute sense to use a GUI Application as a Project. It's more generic and I'm sure I can bring across the points I wanted to make. Will rewrite tomorrow.

Did you still plan to rewrite this to be a specific project?

One thing I was thinking is that it might be nice to pair the new project with a status quo story that leans on it, or at least the description of one. Do you have any thoughts on that?

Fice:
I actually wrote a status quo story on Saturday before life got in the way. It's now commited so you can have a read.

I did not have any time on writing the actual GUI Project , but #71 seems like a good start with only minor adjustments needed.

I do have an idea for one more status quo idea and then quite a few "How i wished it would work shiny future ones". I hope everything settles down the day after tomorrow so I can write up the project and the second story.

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
+# 😱 Status quo stories: Alan has an external event loop and wants to use futures/streams
```
nikomatsakis, line 1: cc @rust-lang/wg-async-foundations I didn't notice at first that there is a status quo story snuck in here :)

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+## The story
+
+As a first Rust Project, Alan decides to program his own IRC Client.
+
+Since it is Alans first Project in Rust, it is going to be a private one. He is going to use it on is Mac, so he decides to go with the cocoa crate to not have to learn any Framework specific quirks. This way Alan can get a feel of Rust itself.
+
+### Alans hopes and dreams
+Despite a learning curve, he managed to creating a first window and have some buttons and menus works. After the initialisation is done, the App hand over control to [CFRunLoop::Run](https://developer.apple.com/documentation/corefoundation/1542011-cfrunlooprun?language=occ).
+
+Once Alan is happy whit his Mock UI, he wants to hook it up with some actual features. He is happy to learn that Rust has Features he already knows.
```
nikomatsakis, line 18:
```suggestion
Once Alan is happy with his Mock UI, he wants to hook it up with some actual features. He is happy to learn that Rust has Features he already knows.
```

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+```rust
+impl Runtime {
+    fn make_progress() {
+        while let Some(task) = self.awoken_tasks.pop() {
+            task.poll();
+            //... remove finished task from 'tasks'
+        }
+    }
+    fn run() {
+        while !self.tasks.is_empty() {
```
nikomatsakis, line 62: This is a bit confusing to me -- after all, `run()` *does* block the OS event loop.. what is it that differentiates this simple Runtime from the others? 

Also, I would expect he does want this runtime to interoperate with the OS runtime...when does he expect these functions to be called?

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+
+...but only for a very short time. Soon after looking into it, he finds out that he has to deal with ```RawWakerVTable```, ```RawWaker```, ```Pointers```. Worst of all, he has to do that without the safety net of the rust compiler, because this stuff is ```unsafe```.
+
+Reimplementing the OS Event Loop is also not an option he wants to take. See [here](https://developer.apple.com/documentation/appkit/nsapplication)
+>Override run() if you want the app to manage the main event loop differently than it does by default. (This a critical and complex task, however, that you should only attempt with good reason).
+
+
+### The cheap way out
+
+Alan gives up and uses a runtime in a seperate thread from the UI. This means he has to deal with the additional burden of syncing and he has to give up the frictionless use of some of the patterns he is accustomed to by treating UI events as ```Stream<Item = UIEvent>```.
```
nikomatsakis, line 79: Can you say a bit more about what he has to do instead? What is the code he *wanted* to write, and what did he get stuck with? 

I'm guessing here the problem is that `UIEvent` can't be sent to other threads?

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+
+## 🤔 Frequently Asked Questions
+
+
+* **What are the morals of the story?**
+    * Even though you come from a language that has async support, does not mean you are used to selecting und driving a runtime.
+    * It should be possible to integrate runtimes into existing Event loops.
+* **What are the sources for this story?**
+    * The authors own experience working on a GUI Framework (very early stage)
+    * Blog post: [Integrating Qt events into Actix and Rust](https://www.rubdos.be/corona/qt/rust/tokio/actix/2020/05/23/actix-qt.html)
```
nikomatsakis, line 89: Nice :)

nikomatsakis (changes requested):
Gave it a quick read -- I have a few clarifying questions! :)

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+As a first Rust Project, Alan decides to program his own IRC Client.
+
+Since it is Alans first Project in Rust, it is going to be a private one. He is going to use it on is Mac, so he decides to go with the cocoa crate to not have to learn any Framework specific quirks. This way Alan can get a feel of Rust itself.
```
nikomatsakis, line 13:
```suggestion
Since it is Alan's first Project in Rust, it is going to be a private one. He is going to use it on is Mac, so he decides to go with the cocoa crate to not have to learn any Framework specific quirks. This way Alan can get a feel of Rust itself.
```

nikomatsakis: link to cocoa crate?

src/vision/status_quo/alan_has_an_event_loop.md
nikomatsakis:
```suggestion
Once Alan is happy with his Mock UI, he wants to make it actually do something. Reading about async Rust, he sees that several of the concepts there map pretty well to some core Cocoa concepts:
```

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+
+Once Alan is happy whit his Mock UI, he wants to hook it up with some actual features. He is happy to learn that Rust has Features he already knows.
+* Promises => Futures
+* Observables => Streams.
+
+Alan smiles, thinking he knows what and more importantly how to do this.
+
+### First time dealing with runtimes
+
+Unfortunately, coming from frameworks like Angular or Node.js, Alan is not used that he himself is responsible for driving the progress of Async/Streams.
```
nikomatsakis, line 26:
```suggestion
Unfortunately, coming from frameworks like Angular or Node.js, Alan is not used that being responsible for diving the progress of Futures/Streams. 
```

Does he first try to write some code and then find that...nothing happens? Or does he just figure that out from reading stuff I guess?

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+Coming from Single-Threaded Angular development, Alan decides to limit his new App to Single-Threaded. He does not feel like learning about Send/Sync/Mutex as well as struggling with the borrow checker.
+
+On top of that, his App is not doing any heavy calculation so he feels async should be enough to not block the main thread too bad and have a hanging UI.
+
+### Fun time is over
+
+Soon Alan realises that he cannot use any of those runtimes because they all take control of the thread and block. The same as the OS Event loop.
```
nikomatsakis, line 49: "those runtimes" => which ones? Maybe this is like "Alan starts reading about the various runtimes that are available for Rust, but he soon realizes that they all take control of the thread and block, and so none of them are going to work."

nikomatsakis: That said, I think that smol can be used in this sort of way. I'll have to dig deeper, but methods like this might be useful:

https://docs.rs/async-executor/1.4.0/async_executor/struct.Executor.html#method.try_tick

This isn't to invalidate your story! Might just be useful for you though. =)

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+
+Alan spends quite some time to look through several runtime implementations. Ignoring most internal things, all he wants is a runtime that looks a bit like this:
```
nikomatsakis, line 51: I guess the key point is that this will process whatever futures are ready, and that's it? One question I have is how/when the OS would know to wake up and execute this method...

src/vision/status_quo/alan_has_an_event_loop.md
```diff
@@ -0,0 +1,104 @@
...
+    }
+    fn run() {
+        while !self.tasks.is_empty() {
+            self.make_progress();
+        }
+    }
+}
+```
+
+It could be soo easy. Unfortunately he does not find any such solution. Having already looked through quite a bit of low level documentation and runtime code, Alan thinks about implementing his own runtime...
```
nikomatsakis, line 69:
```suggestion
It could be so easy. Unfortunately he does not find any such solution. Having already looked through quite a bit of low level documentation and runtime code, Alan thinks about implementing his own runtime...
```

nikomatsakis (changes requested):
Some more notes -- I'm going to hold off on further review for now =)

nikomatsakis:
@Fice do you think you'll have time to revisit some of the comments and questions here?

nikomatsakis:
If it helps, we could try to schedule a writing session perhaps.

nikomatsakis (approved):

nikomatsakis:
I've still got various questions, but I'm inclined to merge this since @Fice seems to be busy. @Fice -- feel free to open a fresh PR with edits or amendments, or to clarify some of the questions I raised here.

PR #81, opened 2021-03-22
Clarify Statis quo, grammar

nealmcb:


nealmcb:
Thank you.  Rust has such a great community. Not even a word of pushback or irony about my PR on fixing a _typo_, which itself had _two distinct ways of misspelling_ the name of the page I was helping to fix!

PR #85, opened 2021-03-24
status quo: Barbara carefully dismisses embedded `Future`

jrvanwhy:
[Rendered](https://github.com/jrvanwhy/wg-async-foundations/blob/bcdef/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md)

Barbara tries to implement an asynchronous API for embedded systems, and finds that `core::future::Future` has a very high overhead.

### DIFF
--- a/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
+++ b/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
@@ -117,20 +117,28 @@ She then modifies `poll` to set `PRINT_WAKER`:
 `PRINT_WAKER` is stored in `.bss`, which occupies space in RAM but not flash. It
 is two words in size. It points to a
 [`RawWakerVTable`](https://doc.rust-lang.org/stable/core/task/struct.RawWakerVTable.html)
-that is provided by the executor. `RawWakerVTable` does not appear to be
-designed for a no-`alloc` environment. In a no-`alloc` environment, `drop` and
-`clone` are generally no-ops, and `wake`/`wake_by_ref` seem like duplicates.
-Looking at `RawWakerVTable` made Barbara think about the size impact of using
-`Future`.
+that is provided by the executor. `RawWakerVTable`'s design is a compromise that
+supports environments both with and without `alloc`. In no-`alloc` environments,
+`drop` and `clone` are generally no-ops, and `wake`/`wake_by_ref` seem like
+duplicates. Looking at `RawWakerVTable` makes Barbara realize that even though
+`Future` was designed to work in embedded contexts, it may have too much
+overhead for her use case.
 
-Barbara decided the overhead would be worth it if it made it easy to work with
-future combinators from external crates *and* could function as a lightweight
-form of multithreading. To support this, Barbara decides to implement an
-executor designed to function like a background thread. Because `alloc` is not
-available, this executor contains a single future. The executor has a `spawn`
-function that accepts a future and starts running that future (overwriting the
-existing future in the executor if one is already present). Once started, the
-executor runs entirely in kernel callbacks.
+Barbara decides to do some benchmarking. She comes up with a sample application
+-- an app that blinks a led and responds to button presses -- and implements it
+twice. One implementation does not use `Future` at all, the other does. Both
+implementations have two asynchronous interfaces: a timer interface and a GPIO
+interface, as well as an application component that uses the interfaces
+concurrently. In the `Future`-based app, the application component functions
+like a future combinator, as it is a `Future` that is almost always waiting for
+a timer or GPIO future to finish.
+
+To drive the application future, Barbara implements an executor. The executor
+functions like a background thread. Because `alloc` is not available, this
+executor contains a single future. The executor has a `spawn` function that
+accepts a future and starts running that future (overwriting the existing future
+in the executor if one is already present). Once started, the executor runs
+entirely in kernel callbacks.
 
 Barbara identifies several factors that add branching and error handling code to
 the executor:
@@ -152,28 +160,21 @@ the executor:
    duplicates the "ignore spurious wakeups" functionality that exists in the
    future itself.
 
-Ultimately, this made the executor logic [quite
-nontrivial](https://github.com/tock/design-explorations/blob/master/size_comparison/futures/src/task.rs).
-The executor logic can be monomorphized for each future, which allows the
-compiler to make inlining optimizations, but results in a significant amount of
-duplicate code. Alternatively, it could be adapted to use function pointers or
-vtables to avoid the code duplication, but then the compiler *definitely* cannot
-inline `Future::poll` into the kernel callbacks.
-
-At this point, Barbara realizes she needs to do some benchmarking. Barbara comes
-up with a sample application -- an app that blinks a led and responds to button
-presses -- and implements it twice. One implementation does not use `Future` at
-all, the other does. In addition to two async drivers (a timer driver and a GPIO
-driver), the `Future`-based app uses Barbara's executor as well as a single
-combinator (that manages the state of the app).
+Ultimately, this made the [executor
+logic](https://github.com/tock/design-explorations/blob/master/size_comparison/futures/src/task.rs)
+nontrivial, and it compiled into 96 bytes of code. The executor logic is
+monomorphized for each future, which allows the compiler to make inlining
+optimizations, but results in a significant amount of duplicate code.
+Alternatively, it could be adapted to use function pointers or vtables to avoid
+the code duplication, but then the compiler *definitely* cannot inline
+`Future::poll` into the kernel callbacks.
 
 Barbara publishes an
 [analysis](https://github.com/tock/design-explorations/tree/master/size_comparison)
 of the relative sizes of the two app implementations, finding a large percentage
 increase in both code size and RAM usage (note: stack usage was not
-investigated). Most of the code size increase is from the future combinator
-code. There are two kernel callbacks in use (one that notifies the app of a
-button press and one that notifies it of a timer expiration).
+investigated). Most of the code size increase is from the future
+combinator code.
 
 In the no-`Future` version of the app, a kernel callback causes the following:
 
@@ -224,13 +225,13 @@ instead. Here are some ways in which these APIs are lighter weight than a
 * **What are the morals of the story?**
     * `core::future::Future` isn't suitable for every asynchronous API in Rust.
       `Future` has a lot of capabilities, such as the ability to spawn
-      dynamically-allocated futures, that are unnecessary in embedded system.
+      dynamically-allocated futures, that are unnecessary in embedded systems.
       These capabilities have a cost, which is unavoidable without
       backwards-incompatible changes to the trait.
     * We should look at embedded Rust's relationship with `Future` so we don't
-      fragment the embedded Rust ecosystem. Other embedded crates use `Future`,
-      presumably because they are less space constrained than Barbara's
-      codebase.
+      fragment the embedded Rust ecosystem. Other embedded crates use `Future`
+      -- `Future` certainly has a lot of advantages over lighter-weight
+      alternatives, if you have the space to use it.
 * **Why did you choose *Barbara* to tell this story?**
     * This story is about someone who is an experienced systems programmer and
       an experienced Rust developer. All the other characters have "new to Rust"
--- a/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
+++ b/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
@@ -31,9 +31,10 @@ sequence of events for a console print:
 
 Barbara tries to implement the API using
 [`core::future::Future`](https://doc.rust-lang.org/stable/core/future/trait.Future.html)
-so that the library can be compatible with the async Rust ecosystem. To do
-this, she decides to make the buffer printing API return a Future. She starts
-with a skeleton:
+so that the library can be compatible with the async Rust ecosystem. The OS
+kernel does not expose a Future-based interface, so Barbara has to implement
+`Future` by hand rather than using async/await syntax. She starts with a
+skeleton:
 
 ```rust
 /// Passes `buffer` to the kernel, and prints it to the console. Returns a
@@ -113,21 +114,23 @@ She then modifies `poll` to set `PRINT_WAKER`:
     }
 ```
 
-At this point, Barbara starts to think about the size impact of these changes.
 `PRINT_WAKER` is stored in `.bss`, which occupies space in RAM but not flash. It
 is two words in size. It points to a
 [`RawWakerVTable`](https://doc.rust-lang.org/stable/core/task/struct.RawWakerVTable.html)
-that is provided by the executor. Considering that this code is designed to run
-without `alloc`, `RawWakerVTable` seems awfully wasteful. `drop` is a no-op,
-`clone` seems like it will be a no-op, and `wake`/`wake_by_ref` seem like
-duplicates.
+that is provided by the executor. `RawWakerVTable` does not appear to be
+designed for a no-`alloc` environment. In a no-`alloc` environment, `drop` and
+`clone` are generally no-ops, and `wake`/`wake_by_ref` seem like duplicates.
+Looking at `RawWakerVTable` made Barbara think about the size impact of using
+`Future`.
 
-At this point, Barbara thought about the overhead and decided it would be worth
-it if it made it easy to work with future combinators from external crates, and
-could function as a lightweight form of multithreading. To support this, Barbara
-decides to implement an executor designed to function like a background thread.
-This executor is started using a `spawn` function, and otherwise runs entirely
-in kernel callbacks.
+Barbara decided the overhead would be worth it if it made it easy to work with
+future combinators from external crates *and* could function as a lightweight
+form of multithreading. To support this, Barbara decides to implement an
+executor designed to function like a background thread. Because `alloc` is not
+available, this executor contains a single future. The executor has a `spawn`
+function that accepts a future and starts running that future (overwriting the
+existing future in the executor if one is already present). Once started, the
+executor runs entirely in kernel callbacks.
 
 Barbara identifies several factors that add branching and error handling code to
 the executor:
@@ -143,9 +146,11 @@ the executor:
    a future is being polled then the future should be re-polled, even if the
    current `poll` returns `Poll::Pending`. This requires putting a retry loop
    into the executor.
-1. A kernel callback may call `Waker::wake` after its future finishes executing,
-   and the executor must ignore those wakeups. This duplicates the "ignore
-   spurious wakeups" functionality that exists in the future itself.
+1. A kernel callback may call `Waker::wake` after its future returns
+   `Poll::Ready`. After `poll` returns `Poll::Ready`, the executor should not
+   `poll` the future again, so Barbara adds code to ignore those wakeups. This
+   duplicates the "ignore spurious wakeups" functionality that exists in the
+   future itself.
 
 Ultimately, this made the executor logic [quite
 nontrivial](https://github.com/tock/design-explorations/blob/master/size_comparison/futures/src/task.rs).
@@ -166,8 +171,36 @@ Barbara publishes an
 [analysis](https://github.com/tock/design-explorations/tree/master/size_comparison)
 of the relative sizes of the two app implementations, finding a large percentage
 increase in both code size and RAM usage (note: stack usage was not
-investigated). Barbara concludes that **`core::future::Future` is not suitable
-for highly-resource-constrained environments** due to the amount of code and RAM
+investigated). Most of the code size increase is from the future combinator
+code. There are two kernel callbacks in use (one that notifies the app of a
+button press and one that notifies it of a timer expiration).
+
+In the no-`Future` version of the app, a kernel callback causes the following:
+
+1. The kernel callback calls the application logic's event-handling function for
+   the specific event type.
+2. The application handles the event.
+
+The call in step 1 is inlined, so the compiled kernel callback consists only of
+the application's event-handling logic.
+
+In the `Future`-based version of the app, a kernel callback causes the
+following:
+
+1. The kernel callback updates some global state to indicate the event happened.
+2. The kernel callback invokes `Waker::wake`.
+3. `Waker::wake` calls `poll` on the application future.
+4. The application future has to look at the state saved in step 1 to determine
+   what event happened.
+5. The application future handles the event.
+
+LLVM is unable to devirtualize the call in step 2, so the optimizer is unable to
+simplify the above steps. Steps 1-4 only exist in the future-based version of
+the code, and add over 200 bytes of code (note: Barbara believes this could be
+reduced to between 100 and 200 bytes at the expense of execution speed).
+
+Barbara concludes that **`Future` is not suitable for
+highly-resource-constrained environments** due to the amount of code and RAM
 required to implement executors and combinators.
 
 Barbara redesigns the library she is building to use a [different
@@ -175,7 +208,16 @@ concept](https://github.com/tock/design-explorations/tree/master/zst_pointer_asy
 for implementing async APIs in Rust that are much lighter weight. She has moved
 on from `Future` and is refining her [async
 traits](https://github.com/tock/libtock-rs/blob/master/core/platform/src/async_traits.rs)
-instead.
+instead. Here are some ways in which these APIs are lighter weight than a
+`Future` implementation:
+
+1. After monomorphization, kernel callbacks directly call application code. This
+   allows the application code to be inlined into the kernel callback.
+2. The callback invocation is more precise: these APIs don't make spurious
+   wakeups, so application code does not need to handle spurious wakeups.
+3. The async traits lack an equivalent of `Waker`. Instead, all callbacks are
+   expected to be `'static` (i.e. they modify global state) and passing pointers
+   around is replaced by static dispatch.
 
 ## 🤔 Frequently Asked Questions
 
@@ -183,6 +225,8 @@ instead.
     * `core::future::Future` isn't suitable for every asynchronous API in Rust.
       `Future` has a lot of capabilities, such as the ability to spawn
       dynamically-allocated futures, that are unnecessary in embedded system.
+      These capabilities have a cost, which is unavoidable without
+      backwards-incompatible changes to the trait.
     * We should look at embedded Rust's relationship with `Future` so we don't
       fragment the embedded Rust ecosystem. Other embedded crates use `Future`,
       presumably because they are less space constrained than Barbara's
--- a/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
+++ b/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
@@ -191,7 +191,6 @@ instead.
     * This story is about someone who is an experienced systems programmer and
       an experienced Rust developer. All the other characters have "new to Rust"
       or "new to programming" as a key characteristic.
-    * [Grace] could as worked as well.
  **How would this story have played out differently for the other characters?**
     * [Alan] would have found the `#![no_std]` crate ecosystem lacking async
       support. He would have moved forward with a `Future`-based implementation,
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,211 @@
...
+      dynamically-allocated futures, that are unnecessary in embedded system.
+    * We should look at embedded Rust's relationship with `Future` so we don't
+      fragment the embedded Rust ecosystem. Other embedded crates use `Future`,
+      presumably because they are less space constrained than Barbara's
+      codebase.
+* **Why did you choose *Barbara* to tell this story?**
+    * This story is about someone who is an experienced systems programmer and
+      an experienced Rust developer. All the other characters have "new to Rust"
+      or "new to programming" as a key characteristic.
+    * [Grace] could as worked as well.
```
alevy, line 194: Grace appears twice in this list, I believe this first is the superfluous one.

jrvanwhy: There is a weird juxtaposition between the fact that I feel Grace would have worked for the story as well and the fact I said the story may play out slightly differently with a Grace. Is that what you are referring to when you say that Grace appears twice?

I'm not quite sure how to fix it. It may actually be better to remove the statement that it would have played out differently with Grace, as I'm not sure that Grace's inexperience with Rust would have made a difference to how things played out.

alevy: Ah, yes that was my confusion. To my reading, this first one is not necessary (sure, you _could_ have used Grace, but the explanation for why it would have gone differently sufficiently explains why Grace would have been a poorer choice). But idk :)

jrvanwhy: I thought about it for another while, and removed the statement that Grace would have sufficed as well. I'm less confident in that statement than I am in my statement on how it may have played out differently with Grace.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,210 @@
...
+[`core::future::Future`](https://doc.rust-lang.org/stable/core/future/trait.Future.html)
+so that the library can be compatible with the async Rust ecosystem. To do
+this, she decides to make the buffer printing API return a Future. She starts
+with a skeleton:
+
+```rust
+/// Passes `buffer` to the kernel, and prints it to the console. Returns a
+/// future that returns `buffer` when the print is complete. The caller must
+/// call kernel_ready_for_callbacks() when it is ready for the future to return. 
+fn print_buffer(buffer: &'static mut [u8]) -> PrintFuture {
```
rylev, line 42: Is the reason Barbara did not use async/await notation relevant for this story?

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,210 @@
...
+        Poll::Pending
+    }
+```
+
+At this point, Barbara starts to think about the size impact of these changes.
+`PRINT_WAKER` is stored in `.bss`, which occupies space in RAM but not flash. It
+is two words in size. It points to a
+[`RawWakerVTable`](https://doc.rust-lang.org/stable/core/task/struct.RawWakerVTable.html)
+that is provided by the executor. Considering that this code is designed to run
+without `alloc`, `RawWakerVTable` seems awfully wasteful. `drop` is a no-op,
```
rylev, line 121: It would be great to talk about the impact of this waste. `RawWakerVTable` is only 4 words big. Probably many reading this story won't understand why 4 words is considered wasteful.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,210 @@
...
+decides to implement an executor designed to function like a background thread.
+This executor is started using a `spawn` function, and otherwise runs entirely
+in kernel callbacks.
+
+Barbara identifies several factors that add branching and error handling code to
+the executor:
+
+1. `spawn` should be a safe function, because it is called by high-level
+   application code. However, that means it can be called by the future it
+   contains. If handled naively, this would result in dropping the future while
```
rylev, line 137: I'm not sure I fully grok this point. Can you expand a bit more? Why would the future potentially be dropped while `spawn` executes?

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,210 @@
...
+      or "new to programming" as a key characteristic.
+ **How would this story have played out differently for the other characters?**
+    * [Alan] would have found the `#![no_std]` crate ecosystem lacking async
+      support. He would have moved forward with a `Future`-based implementation,
+      unaware of its impact on code size and RAM usage.
+    * [Grace] would have handled the issue similarly to Barbara, but may not
+      have tried as hard to use `Future`. Barbara has been paying attention to
+      Rust long enough to know how significant the `Future` trait is in the Rust
+      community and ecosystem.
+    * [Niklaus] would really have struggled. If he asked for help, he probably
```
rylev, line 202: This strikes me as an important point. It's fairly "common knowledge" that `Future`s themselves are very lightweight. This story makes it clear that for some resource constrained environments, the machinery required for getting futures to work can sometimes be too heavy weight. I can imagine there being many who would dismiss Niklaus, saying that `Future`s are lightweight and totally appropriate for embedded development without realizing that this might not always be true.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,210 @@
...
+Barbara publishes an
+[analysis](https://github.com/tock/design-explorations/tree/master/size_comparison)
+of the relative sizes of the two app implementations, finding a large percentage
+increase in both code size and RAM usage (note: stack usage was not
+investigated). Barbara concludes that **`core::future::Future` is not suitable
+for highly-resource-constrained environments** due to the amount of code and RAM
+required to implement executors and combinators.
+
+Barbara redesigns the library she is building to use a [different
+concept](https://github.com/tock/design-explorations/tree/master/zst_pointer_async)
```
rylev, line 174: Can you talk more about what design decisions this implementation made that made it more light-weight?

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,210 @@
...
+   it executes. Barbara adds runtime checks to identify this situation.
+1. `Waker` is `Sync`, so on a multithreaded system, a future could give another
+   thread access to its `Waker` and the other thread could wake it up. This
+   could happen while the `poll` is executing, before `poll` returns
+   `Poll::Pending`. Therefore, Barbara concludes that if `wake` is called while
+   a future is being polled then the future should be re-polled, even if the
+   current `poll` returns `Poll::Pending`. This requires putting a retry loop
+   into the executor.
+1. A kernel callback may call `Waker::wake` after its future finishes executing,
+   and the executor must ignore those wakeups. This duplicates the "ignore
```
rylev, line 147: Can you add a sentence on *why* must the executor ignore the spurious wakeups?

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,210 @@
...
+on from `Future` and is refining her [async
+traits](https://github.com/tock/libtock-rs/blob/master/core/platform/src/async_traits.rs)
+instead.
+
+## 🤔 Frequently Asked Questions
+
+* **What are the morals of the story?**
+    * `core::future::Future` isn't suitable for every asynchronous API in Rust.
+      `Future` has a lot of capabilities, such as the ability to spawn
+      dynamically-allocated futures, that are unnecessary in embedded system.
```
rylev, line 185: Is it `Future` itself or rather the wake based API? Also, are we sure that the tradeoffs in code size experienced here are fundamental to getting `Future` to work? Or is it possible that there can be tweaks to the design that will allow these tradeoffs to go away?

rylev (changes requested):
This is *really* insightful - thank you @jrvanwhy! There are some things that can be expanded upon to make this right for merging, but this is well on its way!

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
alevy: (chiming in from Tock here, so ignore if not useful) @jrvanwhy: did you mean that the stable itself (which is 4 words in flash) is wasteful or the 2 words in memory---seems wasteful because without alloc we actually could know statically exactly what concrete type it would be, and also because we need these 2 words for _each_ instance of async, rather than once for any program that uses futures?

Maybe both?

My interpretation had been the 2 words in memory, but @rylev's comment highlighted to me that I actually am not sure which this is referring to.

jrvanwhy: If there's a single copy of `RawWakerVTable` in flash, that's not bad. If there's 1 copy in RAM for each executor instance, that's pretty significant.

When I wrote the executor, I tried to avoid dynamic dispatch so the compiler could perform inlining. This meant the executor depended on the future's type, like:

```rust
pub struct Task<F: Future> {
    future: Option<F>,
}
```

Because of this, I needed the RawWakerVTable to be specific to the type of Future. I implemented this by putting it directly into the executor, i.e.:

```rust
pub struct Task<F: Future> {
    future: Option<F>,
    vtable: RawWakerVTable,
}
```

As a result, I had a `RawWakerVTable` instance in RAM for each `Task`.

Because `future` needs to be mutable, the `Task` gets placed in `.data` in RAM. As a result, the vtable ends up in RAM, and there's one per `Task`.

In retrospect, I could have broken the executor into two parts:

```rust
// Instantiated by client code, lives in flash.
pub struct TaskVTable<F: Future> {
    vtable: RawWakerVTable
}

// Instantiated by client code, lives in RAM
pub struct Task<F: future> {
    future: Option<F>,
    vtable: &'static TaskVTable<F>,
}
```

This would reduce the RAM overhead to 1 word per `Task` and move the overall vtable into RAM. I expect you could do the same with future combinators (although most combinators probably don't need their own `RawWakerVTable` anyway).

If you use the `Locator` trait from my async API design, we could remove the `vtable` member from `Task` and replace it with a zero-sized type.

nikomatsakis (approved):
This was great. I think @rylev identified everything that could use improvement. Thanks for writing this.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: Yes, I think it is. My understanding is that async/await notation can only block on other futures, and that at the leaf nodes where Rust's async APIs meet an external async API you need handwritten futures. In this case, this driver was interacting directly with the kernel.

I added a sentence to the above paragraph ("Barbara tries to implement... skeleton") explaining this.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: `spawn` takes the future to be executed as an argument. Because the executor is designed to work without `alloc`, it can only hold a single future inside. If `spawn` is called while a future is running I assume that means the caller wants to overwrite the existing future.

I added a note a couple paragraphs up (just before "Barbara identifies several factors...") with an explanation.

I could avoid this cost by making the future fixed (i.e. construct it in a `const` context) at the expense of flexibility.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I added 2 things to address this:

1. An analysis of why the `Future`-based app is larger than the no-`Future` app (currently lines 174-200)
2. A list below (currently lines 214-220) explaining how I (Barbara) avoided the same costs in my async traits.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I didn't really distinguish `Future` from `Waker`.

I think there are fundamental tradeoffs at play. The fact that `Future` is demand-driven results in an information loss during callbacks. Kernel callbacks need to stash data somewhere for `poll` to retrieve later, and `poll` has to contain that retrieval logic. That ended up being the majority of the cost.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: In 2016, Aaron Turon described `Future` as [zero cost](https://aturon.github.io/blog/2016/08/11/futures/#zero-cost), but he elaborated on what that meant, and "minimal impact on code size and RAM usage" was not a part of that definition. The phrase "zero cost" has been repeated in descriptions of `Future` ever since, generally without the context that Aaron provided.

Most of Rust's "zero cost" abstractions have a much smaller (or zero) impact on code size and RAM usage after inlining, so it's easy to assume that `Future` is lighter weight than it is.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I reworded it to state that the executor should not `poll` a future after the future has returned `Poll::Ready`, which is directly stated in the `Future` docs. Is it clear now?

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,254 @@
...
+executor runs entirely in kernel callbacks.
+
+Barbara identifies several factors that add branching and error handling code to
+the executor:
+
+1. `spawn` should be a safe function, because it is called by high-level
+   application code. However, that means it can be called by the future it
+   contains. If handled naively, this would result in dropping the future while
+   it executes. Barbara adds runtime checks to identify this situation.
+1. `Waker` is `Sync`, so on a multithreaded system, a future could give another
```
Matthias247, line 142: As outlined earlier, this is some kind of semi-legit design. The `Waker`s really should just schedule the application tasks again and not trying to poll inline.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,254 @@
...
+   a future is being polled then the future should be re-polled, even if the
+   current `poll` returns `Poll::Pending`. This requires putting a retry loop
+   into the executor.
+1. A kernel callback may call `Waker::wake` after its future returns
+   `Poll::Ready`. After `poll` returns `Poll::Ready`, the executor should not
+   `poll` the future again, so Barbara adds code to ignore those wakeups. This
+   duplicates the "ignore spurious wakeups" functionality that exists in the
+   future itself.
+
+Ultimately, this made the executor logic [quite
```
Matthias247, line 155: I don't think less than 100 LoC with lots of comments is "quite nontrivial". Certainly concurrent code is never easy. But for an executor in general this is fairly simple.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,254 @@
...
+In the `Future`-based version of the app, a kernel callback causes the
+following:
+
+1. The kernel callback updates some global state to indicate the event happened.
+2. The kernel callback invokes `Waker::wake`.
+3. `Waker::wake` calls `poll` on the application future.
+4. The application future has to look at the state saved in step 1 to determine
+   what event happened.
+5. The application future handles the event.
+
```
Matthias247, line 196: But the `Future`s based version provides a lot more guarantees! E.g. it avoids reentrancy problems and makes sure all state associated with the async operation is only modified in a single place (and from a single thread/task). It's a bit of an apples to oranges comparison. If a trivial callback is good enough then `Future`s will likely have overhead. But the places where state has to be persisted in a complex fashion across  `.await` pointers is where `Future`s win, since they don't require encoding of complex state machines.

Matthias247 (commented):
I'm not sure what to make out of this. I think there is a lot of truth that `Future`s  are not zero-cost and it might impact embedded development. Some things that came to my mind here are:
- The lazy nature of `Future`s require them to be constructed on the stack before being moved to a place where they are executed. Huge `Future`s on the stack had been observed. That could easily blow the memory limits of a constrained device.
- Building an allocator for variably sized `Future`s to support spawn might be tricky (if `spawn` only at startup is good enough could probably be done with a bump allocator, but that requires a custom `Box` and allocator APIs are also not fully there yet)
- `Waker`s can outlive `Future`s, and executors need to support the `.wake()` of a `Future` which no longer exists. But reference-counting is not an option, and static state is a bit tricky in combination with the `Send/Sync` requirements.

But the experience report here centers a lot more on what had been experienced on a design which is centered around tock-os. In other places polling `Future`s from arbitrary contexts would likely not be safe.

Maybe this report could provide some input on where documentation on how to build and use `Future`s and executors could be improved?

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I'm having trouble understanding you here. Are you suggesting something like the following?

1. `wake` marks a `Future` as ready to run, but does not call `poll`.
2. The executor offers a `poll_future` function that polls the future if the future has been awoken (repeatedly, if necessary).
3. The application code has to call `poll_future` after invoking kernel callbacks.

Then the application would need to have a main loop, and that loop would look something like the following:

```rust
loop {
    TASK1.poll_future();
    TASK2.poll_future();
    TASK3.poll_future();
    kernel_ready_for_callbacks();
}
```

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: Avoiding reentrancy problems, making async state management easy, and easily representing complex asynchronous operations are all *really* nice features of `Future`! I don't contest that.

However, they are all nice-to-have features, and it takes an awful lot of nice-to-have features to outweigh one deal-breaker. The amount of code generated by the executor and in `poll` that doesn't exist in the non-`Future` version is probably a deal-breaker for us.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
Matthias247: Yes. But the main-loop would typically be something like `executor.run()` - and you've spawned all your tasks on the executor upfront.

Matthias247: After all something like a Winapi/QT/GTK eventloop are also executors :-)

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,254 @@
...
+   wakeups, so application code does not need to handle spurious wakeups.
+3. The async traits lack an equivalent of `Waker`. Instead, all callbacks are
+   expected to be `'static` (i.e. they modify global state) and passing pointers
+   around is replaced by static dispatch.
+
+## 🤔 Frequently Asked Questions
+
+* **What are the morals of the story?**
+    * `core::future::Future` isn't suitable for every asynchronous API in Rust.
+      `Future` has a lot of capabilities, such as the ability to spawn
```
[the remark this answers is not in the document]
jrvanwhy: You can have non-static `Future`s, but you cannot have non-static asynchronous operations with the traits I designed. I think the ability to support non-static futures is a feature of `Future`.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,254 @@
...
+        }
+        unsafe { PRINT_WAKER = Some(cx.waker()); }
+        Poll::Pending
+    }
+```
+
+`PRINT_WAKER` is stored in `.bss`, which occupies space in RAM but not flash. It
+is two words in size. It points to a
+[`RawWakerVTable`](https://doc.rust-lang.org/stable/core/task/struct.RawWakerVTable.html)
+that is provided by the executor. `RawWakerVTable` does not appear to be
```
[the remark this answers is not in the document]
jrvanwhy: > The whole `RawWaker` mechansim was **especially** designed for supporting no-std environments.

I reworded and reorganized this segment significantly, because I definitely didn't convey my thoughts accurately. `RawWaker` was designed to work in both `alloc` and no-`alloc` environments, and it made compromises to support both. I decided to evaluate the overhead of using `Future` when I noticed the compromises in `RawWaker`'s design in order to see if other compromises cost more.

Ultimately, the 2 extra words in `RawWaker` are not a deal-breaker.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,254 @@
...
+`Future`.
+
+Barbara decided the overhead would be worth it if it made it easy to work with
+future combinators from external crates *and* could function as a lightweight
+form of multithreading. To support this, Barbara decides to implement an
+executor designed to function like a background thread. Because `alloc` is not
+available, this executor contains a single future. The executor has a `spawn`
+function that accepts a future and starts running that future (overwriting the
+existing future in the executor if one is already present). Once started, the
+executor runs entirely in kernel callbacks.
```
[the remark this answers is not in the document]
jrvanwhy: > In general `Future`s assume they are only called by one executor in a serial fashion, and not from a variety of places. Doing the latter would void all `Send/Sync` guarantees.

Everything happens in a single thread, so I don't see how `Send`/`Sync` are relevant.

> And not only this, it would open up the door to a variety of reentrancy problems that the Future design avoids.

The executor described here will not call `poll` reentrantly. I don't see anywhere else reentrancy problems can occur.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I see how that would work, but the benefit is not clear to me.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I rephrased this to be more accurate. What I meant by "quite nontrivial" is it generated 96 bytes of code.

jrvanwhy:
> Maybe this report could provide some input on where documentation on how to build and use `Future`s and executors could be improved?

Building a `Future` by hand seems easy enough (except for combinators), but I think there is room for better documentation for building executors.

It's been over a year since I ran this experiment, so I'm unsure what the documentation looked like at the time. However, searching for documentation on building executors gives a handful of tutorials, all of which require `alloc` (even with "embedded" added to my search). A no-`alloc` executor ends up looking significantly different from the executors in the tutorials.

Ultimately, I realized I had to handle the case where `Waker::wake()` is called during `poll` by thinking about what happens in a multithreaded context. I did not find that requirement in any of the documentation I looked at.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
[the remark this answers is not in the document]
Matthias247: One example is that a `Future` which is `!Send` can expect all state which is stored in thread-local storage to be consistently accessible after polls. If one calls a future from arbitrary threads that is no longer true. In general an interrupt context is similar to a separate thread. I'm not familiar enough with tock to say whether this applies or doesn't apply there.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I suppose that you could make all `RawWaker` methods no-ops and instead poll every unfinished `Future` on every iteration, saving some code in the executor at the cost of runtime performance.

Matthias247: This is e.g. required to avoid the reentrancy problems. Let's say our async task calls a UART driver, which submits some bytes. It only writes half of it as part of an `async fn write(&self, &[u8]) -> Result<usize, Error>`. The amounts of bytes written is signalled by the interrupt context, and then forwarded as the result of this sub-`Future` - which returns `Ready`. From there another call to the UART driver is made to start another async operation. This operation is now started from an interrupt context instead of the normal application context. In usual environments that is not allowed, because it now requires that resource to be reentrancy safe. 

When the usual `Future` model is implemented it isn't required to be reentrancy safe, because the execution model already guarantees that from inside a callback no further operations are started. The only thing which will ever happen is waking up the task in the executor, so that it can later run from a safe context. 

This also prevents some other issues, e.g. "what happens if this resource is freed inside the callback". But maybe that's not too important in your "everything is static and will never get freed" scenario.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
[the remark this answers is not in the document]
jrvanwhy: > In general an interrupt context is similar to a separate thread. I'm not familiar enough with tock to say whether this applies or doesn't apply there.

Oh, that doesn't apply in Tock. Invoking a callback behaves almost identically to a virtual function call whose address is determined by the kernel.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
Matthias247: > I suppose that you could make all RawWaker methods no-ops and instead poll every unfinished Future on every iteration, saving some code in the executor at the cost of runtime performance.

Yes, that is indeed the most basic executor. 

More efficient versions can e.g. block the main thread using an OS synchronization mechanism or a low level one like `WFI`, and wake the executor from the interrupt.

https://docs.rs/embedded-executor/0.6.1/embedded_executor/ seems like a decent example on what can be done.

Matthias247: Regarding the dynamic allocs there: You would avoid them by making sure they only run at startup, and size the memory pool which is used for it accordingly (yes, requires some fiddling :'( ). For the list of ready tasks, it's probably doable with a fixed size list or even some bitmasks if a well known amount of tasks need to be executed.

nikomatsakis:
One quick note as I read some of the prior comments. Remember that the role of these status quo stories is for people to **convey what they experienced**. It is explicitly **not** to propose solutions. Therefore, e.g. when @Matthias247 suggests this:

> Maybe this report could provide some input on where documentation on how to build and use Futures and executors could be improved?

I would actually not want that at all. I think the report should focus on what people felt and experienced. 

The stories should definitely be technically accurate -- but in some cases, characters thought things that were wrong, or only partially informed. We should be able to capture that, and the FAQ can be especially useful for providing deeper technical info than the character knew at the time.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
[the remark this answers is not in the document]
nikomatsakis: Ah, this is a good call out @jrvanwhy -- I've been looking for "design decisions" we made along the way that may inform the overall async tenets. There was definitely a push to remove `Send+Sync+'static` bounds from future.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
nikomatsakis: This was a good change. I would have interpreted "quite nontrivial" differently as well.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+So far so good. Barbara then works on `poll`:
+
+```rust
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        if kernel_is_print_done() {
+            Poll::Ready(kernel_get_buffer_back())
+        }
+        Poll::Pending
```
gilescope, line 84: else or `return` missing here?

jrvanwhy: Yes, there was supposed to be a `return`. Fixed.

tmandry:
Thanks so much for writing this up, @jrvanwhy! I'll second @nikomatsakis that I think this can be merged.

nikomatsakis:
One thing that we were discussing on zulip is that some of the comments on this PR might make good FAQs.

jrvanwhy:
> One thing that we were discussing on zulip is that some of the comments on this PR might make good FAQs.

I'll send the changes in a separate PR.

Issue #59, opened 2021-03-19
dropping database handles is hard


nikomatsakis:
* Brief summary: Barbara tried to use sqlx in [YouBuy], but she has problems around cancellation. Ultimately it comes down to the fact that, in `Drop`, the cleanup code to close the database handle is running in a spawned task, and there is a race condition between that task executing and other connections opening.
* Character: [Barbara], could kind of be anyone
* Key points or morals (if known):
    * @galich told me this story

[cast]: https://rust-lang.github.io/wg-async-foundations/vision/characters.html
[Barbara]: https://rust-lang.github.io/wg-async-foundations/vision/characters/barbara.html
[YouBuy]: https://rust-lang.github.io/wg-async-foundations/vision/projects/YouBuy.html

galich:
A real life example: an app is using SQLite and runs a number of separate async tasks, some of them needs to write to the database. SQLite is a single user database that locks database when write is being done. Tasks seems to coordinate well with each other, but sometimes some random task fails to open database because "another process opened it".

After some fun investigation we nailed it to the sqlx crate dropping connection by spawning another async task. We expected connection and associated resources to be free after drop, but in reality it was freed with some lag. Sometimes it could interfere with the rest of the system.

nikomatsakis:
cc @pcwalton and @nagisa -- re: async drop

nikomatsakis:
Part of this blog post feels (maybe) related:

https://kevinhoffman.medium.com/rust-async-and-the-terrible-horrible-no-good-very-bad-day-348ebc836274

>     Lesson III — If you drop (or allow to be dropped) the tokio runtime, any pending tasks are cancelled.

nagisa:
s/database// in general. Here's another very similar instance I recently have seen:

Barbara is implementing SLOW over a cloud messaging service. On the receiving side the messages are accompanied with a handle that allows code to acknowledge or not-acknowledge the message, as well as extend the timeout after which the message would automatically be considered to not-acknowledged. Barbara would like the library to enforce that the messages are always ACKd or NACKd, by sending a NACK when user does not say otherwise. Today Barbara is forced to rely on `#[must_use]` and users not setting timeouts too far into the future to emulate the desired behaviour.

Moral: story around resource cleanup is not quite complete in async Rust;
Source: Efforts to add a Subscriber API to the [hedwig](https://github.com/standard-ai/hedwig-rust) library;
[...]

PR #89, opened 2021-03-25
Status quo: Alan finds use of sqlx bad fit due to db handles

doc-jones:
Alan tries to use sqlx but finds that not being able to explicitly call Drop::drop leads to intermittent errors when connecting to a database in response to an http request. 

Fixes https://github.com/rust-lang/wg-async-foundations/issues/59


nikomatsakis:
cc @galich -- we took a stab at writing up a variant of your story :)

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,84 @@
...
+
+Alan sees the reports coming in from his beta users, but hasn't received any alarms from his application. He realizes he must start debugging. His initial testing produces the customized page for the current user as expected, but after a number of attempts he sees the page with the error message. His application shows an error that says that there is already an open database handle.
+
+
+## Searching for the Solution
+
+
+Alan tries to figure out what happened from the logs, but is unable to do so. Alan decides to attempt to recreate the scenario by doing extra load testing and trying to emulate the increased traffic. After some time, Alan gets lucky and is able to recreate the error, but he still doesn't know what could be the cause. Alan turns to the documentation for the `sqlx` crate to see if there are flags that might enable extra instrumentation but he can't find any (XXX check what crate offers).
+
+Alan, next, takes the following steps to determine the most likely source to explain the problem behavior in his application:
```
nikomatsakis, line 42: @galich in particular, we were trying to imagine how you debugged the problem -- what actually happened, if you recall?

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,84 @@
...
+
+1. Check the `sqlx` documentation for possible flags to explicitly close the connection but none are found.
+2. Examine the source code to figure out where the connection gets closed.
+3. Discover that the connection is closed through the destructor, Drop::drop only when it gets cleaned up aftering going out of scope.
+
+Alan realizes he must seek help from someone more experienced and knowledgable. He goes to his friend Barbara, and asks, "How can I get a log statement when the destructor runs?". Barbara tells him to wrap his database handle in another struct which has a destructor and to insert a log into the destructor for that struct we'll call `Foo`. Alan implements `Foo` and sees the new log when the destructor runs in the output. Now that the log works Alan runs program with full load in order to reproduce the problem. Alan analyzes the logs and notices that the new connection is being created and panicing before the destructor has run.
```
nikomatsakis, line 48: it might be nice to have code examples, I imagine something like

```rust
struct MyConnection {
    connection: SqliteConnection
}

impl Drop for MyConnection {
    fn drop(&mut self) {
        log("dropping the connection");
    }
}
```

But I just realized, this won't actually work! this will print right away, but the *connection itself* doesn't get dropped when the dtor runs, that's kind of the whole point!

nikomatsakis: we might have to weak this part of the story.

nikomatsakis (changes requested):

gilescope:
[Rendered](https://github.com/doc-jones/wg-async-foundations/blob/master/src/vision/status_quo/alan_finds_database_drops_hard.md)

gilescope:
As an alan reading this I wondered how the panic causing the trouble was missed. But I am also worried that destructors don't get run - in all circumstances other than a panic would the destructor get run eventually assuming one of the standard executors?

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Finding the Solution
+
+Alan briefly considers rearchitecting his application in more extreme ways to retain use of async, but he gives up and seeks a more straight forward solution. He discovers `rusqlite`, a sychronous database library and adopts it. This requires some rearchitecting but solves the problem.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+* **What are the morals of the story?**
```
rylev, line 68: I just want to note here that the FAQ section should be filled out before we merge.

nikomatsakis:
> As an alan reading this I wondered how the panic causing the trouble was missed. But I am also worried that destructors don't get run - in all circumstances other than a panic would the destructor get run eventually assuming one of the standard executors?

The problem here probably needs to be spelled out more. It's not that the destructor is not run. What happens is:

* the sqlx connection gets dropped; this needs to close the connection, but that is an async operation, and Drop is a sync context;
* to manage this, [sqlx spawns a task to do the drop](https://github.com/launchbadge/sqlx/blob/0ed524d65c2a3ee2e2a6706910b85bf2bb72115f/sqlx-core/src/pool/connection.rs#L70-L74);
* but this doesn't execute immediately, and hence it's possible to try and create another connection before that spawned task has completed;
* this new connection can fail.

src/vision/status_quo/alan_finds_database_drops_hard.md
doc-jones: Yes, that was understood, but I didn't want to add FAQs until other questions within the story were resolved.  Once the story is complete it will be easier to add FAQs that correctly reflect the details in this story.

nikomatsakis:
@doc-jones and I spent some time revising this story and I believe they are going to post an updated copy.

src/vision/status_quo/alan_finds_database_drops_hard.md
galich: In our case debugging was a bit painful because the system is fairly large, with many moving parts and a lot of async code. So initially we went on a few false leads, thinking that there is a problem of how we line up futures for execution or that there is a rogue process touching database or there is a migration running, etc.

Eventually we thought that we are not closing connection properly, so we double checked that we do drop before another task is trying to access it. When all was verified we decided to look into the fact that sqlx is based on async itself and started to look into the code, eventually discovering that some drops spawn new futures to actually close connections. That was totally unexpected behaviour.

At some point in time it became really strange, if I remember correctly, we saw that there are async spawns inside drops in sqlx, but I think sqlite was not using them. That was confusing, but then we remembered that we use connection pooling which has some async spawning in drop handler https://github.com/launchbadge/sqlx/blob/0ed524d65c2a3ee2e2a6706910b85bf2bb72115f/sqlx-core/src/pool/connection.rs#L74 .

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+Alan tries to figure out what happened from the logs, but is unable to do so. Alan decides to attempt to recreate the scenario by doing extra load testing and trying to emulate the increased traffic. After some time, Alan gets lucky and is able to recreate the error, but he still doesn't know what could be the cause. Alan turns to the documentation for the `sqlx` crate to see if there are flags that might enable extra instrumentation but he can't find any (XXX check what crate offers).
+
+Alan, next, takes the following steps to determine the most likely source to explain the problem behavior in his application:
+
+1. Check the `sqlx` documentation for possible flags to explicitly close the connection but none are found.
+2. Examine the source code to figure out where the connection gets closed.
+3. Discover that the connection is closed through the destructor, Drop::drop only when it gets cleaned up aftering going out of scope.
+
```
galich, line 47: > 1. Check the `sqlx` documentation for possible flags to explicitly close the connection but none are found.

we were more likely to believe in error in our code than in a crate

> 2. Examine the source code to figure out where the connection gets closed.

most of the times closing is automatic, so we had expectations that once it's dropped all resources are released

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+Things seem to be working fairly well but sometimes when he refreshes the page he encounters a panic with the message "Cannot open a new connecton: connection is already open". He is flummoxed.
+
+
+## Searching for the Solution
+
+
+Alan tries to figure out what happened from the logs, but the only information he sees is that a new connection has been received. Alan turns to the documentation for the `sqlx` crate to see if there are flags that might enable extra instrumentation but he can't find any [sqlx::Connection::close](https://docs.rs/sqlx/0.5.1/sqlx/trait.Connection.html#required-methods).
```
nikomatsakis, line 40:
```suggestion
Alan tries to figure out what happened from the logs, but the only information he sees is that a new connection has been received. Alan turns to the documentation for the `sqlx` crate to see if there are flags that might enable extra instrumentation but he can't find any.
```

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+
+* He does find the [`close` method] which mentions "This method is not required for safe and consistent operation. However, it is recommended to call it instead of letting a connection drop as the database backend will be faster at cleaning up resources."
```
nikomatsakis, line 42:
```suggestion
He's a bit confused, because he's accustomed to having things generally be cleaned up automatically when they get dropped (for example, dropping a [`File`](https://doc.rust-lang.org/std/fs/struct.File.html) will close it). Searching the docs, he sees the [`close`](https://docs.rs/sqlx/0.5.1/sqlx/trait.Connection.html#tymethod.close) method, but the comments confirm that he shouldn't have to call it explicitly: "This method is not required for safe and consistent operation. However, it is recommended to call it instead of letting a connection drop as the database backend will be faster at cleaning up resources." Still, just in case, he decides to add a call to `close` into his code. It does seem to help some, but he is still able to reproduce the problem if he refreshes often enough. Feeling confused, he adds a log statement right before calling `close` to see if it is working:
```

Let's go out of bullet points here, I think.

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+* He adds a call to `close` into his code and it helps some but he is still able to reproduce the problem if he refreshes often enough.
```
nikomatsakis, line 43:
```suggestion
```

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+## Searching for the Solution
+
+
+Alan tries to figure out what happened from the logs, but the only information he sees is that a new connection has been received. Alan turns to the documentation for the `sqlx` crate to see if there are flags that might enable extra instrumentation but he can't find any [sqlx::Connection::close](https://docs.rs/sqlx/0.5.1/sqlx/trait.Connection.html#required-methods).
+
+* He does find the [`close` method] which mentions "This method is not required for safe and consistent operation. However, it is recommended to call it instead of letting a connection drop as the database backend will be faster at cleaning up resources."
+* He adds a call to `close` into his code and it helps some but he is still able to reproduce the problem if he refreshes often enough. 
+* He adds a log statement right before calling `close` to see if it is working:
```
nikomatsakis, line 44:
```suggestion
```

src/vision/status_quo/alan_finds_database_drops_hard.md
nikomatsakis: I added some notes to this effect in my suggestions

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+    
+    // he adds this:
+    log!("closing the connection");
+    conn.close();
+
+    Ok(())
+}
+```
+  
+* He observes that in the cases where he has the problem the log statement never executes.
```
nikomatsakis, line 69: I think we can basically just remove the bullet points here

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+* He asks Barbara for help and she explains how `await` can be canceled and it will the destructors for things that are in scope.
```
nikomatsakis, line 70: I wonder if we can link to something here that explains this-- googling brought up [this gist](https://gist.github.com/Matthias247/ffc0f189742abf6aa41a226fe07398a8) by @Matthias247, which could be good

nikomatsakis:
```suggestion
* He asks Barbara for help and she points him to [this gist](https://gist.github.com/Matthias247/ffc0f189742abf6aa41a226fe07398a8) that explains how `await` can be canceled and it will the destructors for things that are in scope.
```

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+* He reads the [source for the SqliteConnection destructor](https://github.com/launchbadge/sqlx/blob/0ed524d65c2a3ee2e2a6706910b85bf2bb72115f/sqlx-core/src/pool/connection.rs#L70-L74) and finds that destructor spawns a task to actually close the connection.
+* He realizes there is a race condition:
+    * the task may not have actually closed the connection before `do_the_thing` is called a second time
+* He gives up and gets frustrated
```
nikomatsakis, line 74:
```suggestion
* At this point, he is feeling pretty frustrated!
```

nikomatsakis: He didn't quite give up *yet*

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+
+Next, Alan seeks verification and validation of his understanding of the source code from the `sqlx` forum. His advisor states that Rust doesn't have a way to execute async operations in a destructor.
```
nikomatsakis, line 76:
```suggestion
Next, Alan seeks verification and validation of his understanding of the source code from the `sqlx` forum. Someone on the forum explains why the destructor launches a fresh task: Rust doesn't have a way to execute async operations in a destructor.
```

but I wonder if it would be better to remove this and move it into the story above. It seems more likely he would ask on the forum and be directed to the source code than to go digging and find it directly. Doesn't matter too much though.

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+## Finding the Solution
+
+Alan briefly considers rearchitecting his application in more extreme ways to retain use of async, but he gives up and seeks a more straight forward solution. He discovers `rusqlite`, a sychronous database library and adopts it. This requires some rearchitecting but solves the problem.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+* **What are the morals of the story?**
+    * The story demonstrates solid research steps that Alan uses to understand and resolve his problem.
```
nikomatsakis, line 87: I'd say a key moral is that Rust's async story is lacking a way to executing async operations in destructors. Spawning is a workaround, but it can have unexpected side-effects.

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+    * Completion of the [Cancellation and timeouts docs](https://rust-lang.github.io/async-book/06_multiple_futures/01_chapter.html) may have been helpful. It's difficult to know how something absent might have improved the solution search process.
+    * Stating the obvious, with more async experience Alan would shorten his solution search process.
+* **What are the sources for this story?**
+    * Sergey Galich reported this problem, among many others.
```
nikomatsakis, line 91:
```suggestion
    * This specific story describes an actual bug encountered by Sergey Galich at 1Password.
```

src/vision/status_quo/alan_finds_database_drops_hard.md
```diff
@@ -0,0 +1,104 @@
...
+* **Why did you choose Alan to tell this story?**
+    * His experience and understanding of other languages coupled with his desire to apply Rust would likely lead him to try solutions before deeply researching them.
+* **How would this story have played out differently for the other characters?**
+    * It is likely that Nickalas would have reached out to Barbara or possibly the sqlx community sooner. On the other hand Barbara, given her deep experience in Rust wouldn't have selected async in her initial design.
```
nikomatsakis, line 95: I'm not sure this part about Barbara is true. I would've probably tried to use async at first. This seems like a pretty subtle bug that could easily take a while for anyone to understand.

nikomatsakis: I'd probably just say that it might well have played out similarly for everyone.

nikomatsakis (changes requested):

src/vision/status_quo/alan_finds_database_drops_hard.md
nikomatsakis: OK. I think the story is *reasonably* close to this, at least.

nikomatsakis:
> As an alan reading this I wondered how the panic causing the trouble was missed. But I am also worried that destructors don't get run - in all circumstances other than a panic would the destructor get run eventually assuming one of the standard executors?

@gilescope I think we resolved that, do you agree?

src/vision/status_quo/alan_finds_database_drops_hard.md
doc-jones: added

src/vision/status_quo/alan_finds_database_drops_hard.md
doc-jones: change committed

src/vision/status_quo/alan_finds_database_drops_hard.md
doc-jones: changes committed

nikomatsakis (approved):

PR #92, opened 2021-03-25
Status Quo: Grace integrates a C API

ThatGeoGuy:
Grace tries to create an async pipeline with a C-API that provides some limited async-like interfaces. She discovers some assimilation bias from C and C++, but manages to produce some working code for her and her team.

### DIFF
--- a/src/vision/status_quo/grace_wants_to_integrate_c_api.md
+++ b/src/vision/status_quo/grace_wants_to_integrate_c_api.md
@@ -148,9 +148,9 @@ of an equivalent function pointer in C, and slightly laments that Rust cannot ex
       safety they would gain from switching to Rust, or exploring async code in a more rigorous way.
     * The async code is actually much simpler than writing the entire execution loop in C themselves. However, the
       assumption that you would write the entire execution loop is baked into the underlying library which Grace's team
-      cannot rewrite entirely from scratch. This makes it feel weird to work with async code in Rust that formalizes
-      concepts such as wake-ups. Even though Grace's team hasn't done anything particularly egregious by sleeping the
-      active task, it still "feels" unsatisfying coming from other languages where such a pattern might be a code smell.
+      cannot rewrite entirely from scratch. Integrating Rust async code with other languages which might have different
+      mental models can sometimes lead to unidiomatic or unsatisfying code, even if the intent of the code in Rust is
+      clear.
     * Closures and related types can be very hard to write in Rust, and if you are used to being very explicit with your
       types, tricks such as the `impl` trick above for `Stream`s aren't immediately obvious at first glance.
 * **What are the sources for this story?**
--- a/src/vision/status_quo/grace_wants_to_integrate_c_api.md
+++ b/src/vision/status_quo/grace_wants_to_integrate_c_api.md
@@ -1,0 +1,176 @@
+# 😱 Status quo stories: Grace wants to integrate a C-API
+
+[Alan]: ../characters/alan.md
+[Grace]: ../characters/grace.md
+[Niklaus]: ../characters/niklaus.md
+[Barbara]: ../characters/barbara.md
+
+[bindgen]: //docs.rs/bindgen/
+[`stream::unfold`]: //docs.rs/futures/0.1.17/futures/stream/fn.unfold.html
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life
+experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today.
+
+## The story
+
+[Grace] is integrating a camera into an embedded project. Grace has done similar projects before in the past, and has
+even used this particular hardware before. Fortunately, the camera manufacturer provides a library in C to interface
+with the driver.
+
+Grace knows that Rust provides strong memory safety guarantees, and the library provided by the manufacturer sports an
+API that is easy to misuse. In particular, ownership concerns are tricky and Grace and her team have often complained in
+the past that making memory mistakes is very easy and one has to be extremely careful to manage lifetimes. Therefore,
+for this project, Grace opts to start with Rust as many of the pitfalls of the manufacturer's library can be
+automatically caught by embedding the lifetimes into a lightweight wrapper over code bridged into Rust with [bindgen].
+
+Grace's team manages to write a thin Rust wrapper over the manufacturer's library with little complication. This library
+fortunately offers two interfaces for grabbing frames from the camera: a blocking interface that waits for the next
+frame, and a non-blocking interface that polls to check if there are any frames currently available and waiting. Grace
+is tempted to write a callback-based architecture by relying on the blocking interface that waits; however, early the
+next morning the customer comes back and informs her that they are scaling up the system, and that there will now be 5
+cameras instead of 1.
+
+She knows from experience that she cannot rely on having 5 threads blocking just for getting camera frames, because the
+embedded system she is deploying to only has 2 cores total! Her team would be introducing a lot of overhead into the
+system with the continuous context switching of every thread.  Fortunately, Grace notices the similarities between the
+polling interface in the underlying C library and the `Poll` type returned by Rust's `Future` trait. "Surely," she
+thinks, "I can asynchronously interleave polls to each camera over a single thread, and process frames as they become
+available!" Such a thing would be quite difficult in C while guaranteeing memory safety was maintained. However, Grace's
+team has already dodged that bullet thanks to writing a thin wrapper in Rust that manages these tricky lifetimes!
+
+### The first problem: polls and wake-ups
+
+Grace sets out to start writing the pipeline to get frames from the cameras. She realizes that while the polling call
+that the manufacturer provided in their library is similar in nature to a future, it doesn't quite encompass everything.
+In C, one might have to set some kind of heartbeat timer for polling. Grace explains to her team that this heartbeat is
+similar to how the `Waker` object works in a `Future`'s `Context` type, in that it is how often the execution
+environment should re-try the future if the call to `poll` returns `Poll::Pending`.
+
+Something Grace learns to love immediately about Rust is that writing the `Future` here does not require her team to
+write their own execution environment. In fact, the future can be entirely written independent of the execution
+environment. She quickly writes an async method to represent the polling process:
+
+```rust
+/// Gets the next frame from the camera, waiting `retry_after` time until polling again if it fails.
+///
+/// Returns Some(frame) if a frame is found, or None if the camera is disconnected or goes down before a frame is
+/// available.
+async fn next_frame(camera: &Camera, retry_after: Duration) -> Option<Frame> {
+    while camera.is_available() {
+        if let Some(frame) = camera.poll() {
+            return Some(frame);
+        } else {
+            task::sleep_for(retry_after).await;
+        }
+    }
+
+    None
+}
+```
+
+The underlying C API doesn't provide any hooks that can be used to wake the `Waker` object on this future up, so Grace
+and her team decide that it is probably best if they just choose a sufficiently balanced `retry_after` period in which
+to try again. It does feel somewhat unsatisfying, as calling `sleep_for` feels about as hacky as calling
+`std::this_thread::sleep_for` in C++. However, there is no way to directly interoperate with the waker without having a
+separate thread of execution wake it up, and the underlying C library doesn't have any interface offering a notification
+for when that should be. In the end, this is the same kind of code that they would write in C, just without having to
+implement a custom execution loop themselves, so the team decides it is not a total loss.
+
+### The second problem: doing this many times
+
+Doing this a single time is fine, but an end goal of the project is to be able to stream frames from the camera for
+unspecified lengths of time. Grace spends some time searching, and realizes that what she actually wants is a `Stream`
+of some kind, so that her team can write something similar to:
+
+```rust
+let frame_stream = stream_from_camera(camera, Duration::from_millis(5));
+
+while let Some(frame) = frame_stream.next().await {
+    // process frames
+}
+
+println!("Frame stream closed.");
+```
+
+She scours existing crates, in particular looking for one way to transform the above future into a stream that can be
+executed many times. The only available option to transform a future into a series of futures is [`stream::unfold`],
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/grace_wants_to_integrate_c_api.md
```diff
@@ -0,0 +1,176 @@
...
+of an equivalent function pointer in C, and slightly laments that Rust cannot express such as clearly.
+
+## 🤔 Frequently Asked Questions
+
+* **What are the morals of the story?**
+    * Rust was the correct choice for the team across the board thanks to its memory safety and ownership. The
+      underlying C library was just too complex for any single programmer to be able to maintain in their head all at
+      once while also trying to accomplish other tasks.
+    * Evolving requirements meant that the team would have had to either start over in plain C, giving up a lot of the
+      safety they would gain from switching to Rust, or exploring async code in a more rigorous way.
```
rylev, line 148: I'm not sure I see this moral reflected in the story. Can you talk a bit more about what you mean by this and how it relates to the story?

src/vision/status_quo/grace_wants_to_integrate_c_api.md
```diff
@@ -0,0 +1,176 @@
...
+    * The async code is actually much simpler than writing the entire execution loop in C themselves. However, the
```
rylev, line 149: Let me know if I'm off-base here, but it seems like the underlying moral here is that integrating Rust async code with other languages which might have different mental models can sometimes lead to unidiomatic or unsatisfying code. Is that correct?

src/vision/status_quo/grace_wants_to_integrate_c_api.md
```diff
@@ -0,0 +1,176 @@
...
+let frame_stream = stream_from_camera(camera, Duration::from_millis(5));
+
+while let Some(frame) = frame_stream.next().await {
+    // process frames
+}
+
+println!("Frame stream closed.");
+```
+
+She scours existing crates, in particular looking for one way to transform the above future into a stream that can be
```
rylev, line 97: The use of the word "scours" makes me think this wasn't a trivial or pleasant process. Perhaps another moral is that understanding that this use case is best modeled as a stream *and* learning how to turn futures into streams were both not easy for someone with a C/C++ background. What do you think?

src/vision/status_quo/grace_wants_to_integrate_c_api.md
```diff
@@ -0,0 +1,176 @@
...
+
+```rust
+struct StreamState {
+    camera: Camera,
+    retry_after: Duration,
+}
+
+
+
+fn stream_from_camera(camera: Camera, retry_after: Duration) -> Unfold<Frame, ??, ??> {
```
nikomatsakis, line 110: Question (maybe something to add to the FAQ): Where did Grace get `Unfold` from? I guess she saw it in the Rustdoc, is that write?

rylev (commented):
This is looking really great. I think there are some places where we can fine tune the morals to be a bit clearer, but I'm loving what I'm seeing so far.

nikomatsakis (approved):
This was very nicely written. I followed along pretty easily. I also held my breath a bit in suspense -- will Grace be able to use Streams? What a relief to know it worked out this time. =)

src/vision/status_quo/grace_wants_to_integrate_c_api.md
nikomatsakis: Good catch -- I would like to hear a bit more about how she figured out that she wanted streams and found the combinator, because that's not entirely obvious to me either.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
nikomatsakis: Sounds plausible, although I found the "sleep.await" to be .. tolerable. Maybe it'd be nice to have a future combinator or something that's like "poll every so often" just to express more clearly what's going on, though. (Maybe this exists, idk)

src/vision/status_quo/grace_wants_to_integrate_c_api.md
nikomatsakis: (I'm asking because I'm thinking a bit about what might have led her to write `impl Stream` from the get go)

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: Thanks for the feedback! I probably did not expound on this too much in the story. 

In reality, when you try to switch towards a new(er) language, there's always the "oops we hit the first roadblock, maybe this whole idea was a mistake." This sentiment happens a lot, especially when switching away from a technology as tried and tested as C. 

I suppose at the point in the story where requirements evolved (1 camera -> 5 cameras), I've seen hesitation to walk back other ambitious parts of a project in order to conserve time or capture risk. This is a bit nuanced, and probably isn't the same across every org! I fully admit that this "moral" is perhaps ill-placed or not-well-explained in this context. In my own experience, we have had debates around whether it was even worth diving into a bigger solution with async given that the project has shifted. Ultimately, I pushed for continuing in Rust because the benefits applied to memory safety could _also_ be true for async, if we could figure out what our unknown unknowns were.

With that in mind, does this moral detract from the overall story? Is the connection too soft? I'm happy to remove it, I think the story as it stands has a lot already, but I could also see adding a small amount of content to help clarify.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: I would say the assessment above is correct! Spinning on a sleep timer when using threads is actually an easy way to get race conditions in C or C++. This is why `std::condition_variable` exists, to make waiting on a certain condition across threads more clear. 

Sleeping in a future here doesn't have any of the same problems as the above, as race conditions aren't really the concern given the over-arching async model. But yes, the moral is effectively as @rylev described. Given that the wording there is much better, I think I'll go ahead and change this to say:

> The async code is actually much simpler than writing the entire execution loop in C themselves. However, the assumption that you would write the entire execution loop is baked into the underlying library which Grace's team cannot rewrite entirely from scratch. Integrating Rust async code with other languages which might have different mental models can sometimes lead to unidiomatic or unsatisfying code, even if the intent of the code in Rust is clear. 

On the topic of some kind of future combinator: this would be helpful! I don't know if that's over-abstraction though, as it seems like loop & sleep work well! One thing I did _not_ mention was that `task::sleep` is actually in `async_std::task`, not `std::task`. I think given how "basic" this is and how easily it maps back to `std::thread::sleep`, there should at least be an async sleep in `std`, so that the above example is truly independent of crate.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: I think the main difficulty (why I used the word "scours") was that it was not easy to search based off of types. I was originally looking into how to implement the Stream trait while using async calls inside `poll_next`! I have a background that also extends to using streams in Scheme, so I was thinking there about how I'd do it, and that was when I started looking through `futures::stream` for possible combinators or constructors for stream types. 

Eventually after reading through the whole module I found the one combinator that had an example taking in a future, instead of constructing a stream from synchronous code. I didn't put 2 and 2 together with regards to `stream::unfold` in Rust and `stream-unfold` in Scheme (see SRFI 45) until I had already read over the code 😅 

But I would generally agree again with this point:

> Perhaps another moral is that understanding that this use case is best modeled as a stream and learning how to turn futures into streams were both not easy for someone with a C/C++ background. What do you think?

Shall I add it to the morals section?

src/vision/status_quo/grace_wants_to_integrate_c_api.md
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    None
+}
+```
+
+The underlying C API doesn't provide any hooks that can be used to wake the `Waker` object on this future up, so Grace
+and her team decide that it is probably best if they just choose a sufficiently balanced `retry_after` period in which
+to try again. It does feel somewhat unsatisfying, as calling `sleep_for` feels about as hacky as calling
```
eminence, line 75: I sympathize with Grace :)

Just a small personal anecdote:  Tokio as a [tutorial](https://tokio.rs/tokio/tutorial/async) that covers how to write your own custom `Future` that will wait for some interval to elapse.  The future that tokio has you write simply spawns a blocking `sleep` in a new thread.  When I first saw this I felt the same type of "unsatisfaction" that Grace probably felt too.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: Firstly: Thanks for reading it, and thank you for the kind words! I appreciate the feedback.

Yes, I saw it in the type signature for `stream::unfold` in the futures crate. Initially I started writing everything as traits (which is perhaps another story) and using `impl Stream`as a return type in traits is... not currently viable. I eventually abandoned the trait architecture because async functions in traits are also problematic (and then get into the complexity of `Pin`, boxing your futures, etc.).

One thing that is somewhat tangential to this whole story is how traits are very much sub-optimal when working with async. Coming from C++ (and wanting some idea of virtual dispatch) it seems natural to use traits for everything, but that is just something to unlearn from C++, independent of async 🙃

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: Thank you for the feedback! I did not know that tutorial existed, and actually I've mostly stuck to async-std over tokio, so I must have missed a lot of valuable tutorials!

I admit, this is a really weird part of async that feels unsatisfying, but it's actually not that bad with regards to the runtime. At least, it isn't as bad as doing the equivalent with `std::thread::sleep` on a thread (which is usually a sign you want a condition variable!).

eminence: Do you think Grace went looking for a tutorial on how to implement a custom Future, and was disappointed when she didn't find one?   Or do you think Grace read the API docs for Future and they made enough sense to her that she felt comfortable jumping in and trying a custom implementation?

ThatGeoGuy: Certainly the latter in this case. Grace definitely read the docs for the `Future` trait, and then intuitively understood how async-await de-sugars to a custom future. 

Grace may have read [this article on why async trait fns are hard](https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/) which provides a ton of background for intuiting the async-await and how the compiler structures / de-sugars that syntax, but otherwise didn't feel any additional need for tutorials given that plus past experience + the rustdocs. The loop -> sleep cycle seemed natural, albeit did feel unidiomatic or unsatisfying coming from C or C++, where that would have probably been a code-smell.

ThatGeoGuy: On that note, is this worth adding to the story? The blog post is fantastic, and a great resource for people jumping into async from C/C++, but I don't know if it is directly applicable without feeling like I forced it in somehow...

eminence: I don't want to unduly color Grace's experiences here, but maybe adding a small note about this into the story would be good.  We have a lot of stories about users who get stuck in various ways on simple things, so it might be nice to mention that some characters (like Grace here) already have the right experience and intuition to jump directly into things like writing custom futures (even when being relatively new to async rust)

nikomatsakis: I definitely think it'd be good to elaborate on this part.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: Added a note to the morals section in 9e0bd1c5f1a95b732361ddcb898d2cd0decb501c. 

Let me know if that clears it up!

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: See 2eedd5a1af3b923030a9e1ab15fdeb0a7451988e for the above change.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
nikomatsakis: Can you maybe add a note about `Unfold` as a FAQ? Something lke

* Q: How did Grace know to use `Unfold` as the return type in the first place?
* She saw it in the rustdoc for `stream::unfold`.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
nikomatsakis: I think it would be helpful to just add text around this. Something like:

The requirements shifted. Some folks argued for moving away from Rust -- maybe they should stick to the tried and true. But ultimately they decided that the benefits of memory safety were still applicable xxx

src/vision/status_quo/grace_wants_to_integrate_c_api.md
nikomatsakis: It looks like this is resolved.

nikomatsakis (changes requested):
I did a pass -- I have a few minor comments, this looks almost ready to merge to me.

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: I added this to the FAQ, with a link to the rustdoc page in question!

src/vision/status_quo/grace_wants_to_integrate_c_api.md
ThatGeoGuy: 👍🏻 I added a note in text for this in  c5891dee9f2fa9230c2799d66bf75887fc6b8a02. Let me know if it needs any more fine-tuning!

nikomatsakis: Looks good to me!

doc-jones (approved):

doc-jones (approved):

doc-jones (approved):

doc-jones (approved):

PR #93, opened 2021-03-26
Status quo: Alan lost the World!

kmeisthax:
Rendered: https://github.com/kmeisthax/wg-async-foundations/blob/status-quo-ruffle/src/vision/status_quo/alan_lost_the_world.md

This is based off of my personal experience trying to use Rust async in Ruffle, where pretty much all of our code has to hold a bunch of borrows to various globals as well as some extra bits our particular GC requires for memory safety. I tried my best to make this somewhat applicable beyond our own *really wonky* architecture, which works well in sync and breaks horribly in async.

I'm also underestimating how much extra verbosity comes from having to re-establish an update context in this story. If you *really* wanna see what we're doing, you can read some of our actual async code: https://github.com/ruffle-rs/ruffle/blob/master/core/src/loader.rs#L413


src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+Niklaus heard about a project to reimplement a deprecated browser plugin using Rust and WASM. This old technology had the ability to load resources over HTTP; so it makes sense to try and implement that functionality using the Fetch API. He looks up the documentation of `web_sys` and realizes that he needs to take a thing called a `Future` out of the call to `fetch`, do some things to it in Rust, and then send the future off to `spawn_local` as such:
+
+```use wasm_bindgen_futures::spawn_local;
```
rylev, line 13: nit: the beginning of the code should be on the next line. This is the case in all the code snippets.

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+Niklaus heard about a project to reimplement a deprecated browser plugin using Rust and WASM. This old technology had the ability to load resources over HTTP; so it makes sense to try and implement that functionality using the Fetch API. He looks up the documentation of `web_sys` and realizes that he needs to take a thing called a `Future` out of the call to `fetch`, do some things to it in Rust, and then send the future off to `spawn_local` as such:
```
rylev, line 11: Does Niklaus understand what `fetch` and `spawn_local` are?

src/vision/status_quo/niklaus_lost_his_context.md
nikomatsakis, line 13:
```suggestion
```rust
use wasm_bindgen_futures::spawn_local;
```

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+        let url = src.to_string();
+        let request = //still elided
+        let data = window().unwrap().fetch_with_request(&request).await.unwrap().etc.etc.etc;
+        let image = parse_png(data, context);
+        context.display_tree.add_child_node(inside_of, image);
+    });
+}
+```
+
+Well, that stopped compiling. After looking through far too many compiler error messages, Niklaus realizes that the problem is the context itself. You see, whenever you close over some borrowed state, your closure's lifetime is bounded by that borrow. This is reasonable for closures, as there are plenty of contexts where you would pass such a closure and have it called immediately. However, the whole point of futures is that they are *not* called immediately - you hold them inside of a thing called an "executor", which typically needs futures that own their contents.
```
rylev, line 45: It'd be helpful to give a flavor of what the error messages look like.

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+                      //just pretend it's here
+        window().unwrap().fetch_with_request(&request).await;
+        log::error!("It worked");
+    });
+}
+```
+
+Niklaus adds calls to `load_image` where appropriate, sees the message pop up in the console, and figures it's time to now actually do something to that loaded image. At this point, it's important to note that many parts of this project pass around "contexts" - bundles of borrowed state that various parts of the reimplementation are free to alter. (This is necessary because this plugin had multiple scripting runtimes in it.) In synchronous code, this was perfectly fine. However, when we start integrating the function into the player...
+
+```use crate::{Index, Context}
```
nikomatsakis, line 29:
```suggestion
```rust
use crate::{Index, Context}
```

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+
+Niklaus heard about a project to reimplement a deprecated browser plugin using Rust and WASM. This old technology had the ability to load resources over HTTP; so it makes sense to try and implement that functionality using the Fetch API. He looks up the documentation of `web_sys` and realizes that he needs to take a thing called a `Future` out of the call to `fetch`, do some things to it in Rust, and then send the future off to `spawn_local` as such:
+
+```use wasm_bindgen_futures::spawn_local;
+use web_sys::window;
+
+fn load_image(src: &url) {
+    spawn_local(async move {
+        let url = src.to_string();
+        let request = //request construction in Rust is actually pretty verbose
```
nikomatsakis, line 19:
```suggestion
        let request = make_request(url);
```

maybe something like that?

then later...


```rust
fn make_request(...) {
    // request construction is actually pretty verbose!
    // [see here for an example](...)
}
```

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+```use crate::{Index, Context}
+use wasm_bindgen_futures::spawn_local;
+use web_sys::window;
+use crate::parse_png;
+
+fn load_image(src: &url, inside_of: Index<Node>, context: &mut Context<'_>) {
+    spawn_local(async move {
+        let url = src.to_string();
+        let request = //still elided
+        let data = window().unwrap().fetch_with_request(&request).await.unwrap().etc.etc.etc;
```
nikomatsakis, line 38: :grin: this says it all :)

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+Well, that stopped compiling. After looking through far too many compiler error messages, Niklaus realizes that the problem is the context itself. You see, whenever you close over some borrowed state, your closure's lifetime is bounded by that borrow. This is reasonable for closures, as there are plenty of contexts where you would pass such a closure and have it called immediately. However, the whole point of futures is that they are *not* called immediately - you hold them inside of a thing called an "executor", which typically needs futures that own their contents.
+
+At this point, Niklaus is looking at either rewriting a good chunk of the program to not use contexts, or rewriting his own code to sidestep the problem. He chooses the latter, by holding a reference to the thing that produces the context and then updating it after the async portion of the task has concluded:
+
+```use crate::{Index, Player}
+use wasm_bindgen_futures::spawn_local;
+use web_sys::window;
+use crate::parse_png;
+
+fn load_image(src: &url, inside_of: Index<Node>, player: Arc<Mutex<Player>>) {
```
nikomatsakis, line 54: it wasn't clear to me that Niklaus had control of this interface -- I thought `load_image` was somehow coming from other code...

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+      * Owned Async, which can only own parameters
+      * Borrowed Async, which can both own and borrow parameters, but gains a lifetime
+    * Non-`'static` Futures are of limited use as lifetimes are tied to the sync stack. I think you can use `block_on` with them but that's about it.
+    * Borrowed Async functions would be useful if there was a way to have a future whose borrows were scoped to the moment they are `poll`ed.
+* **What are the sources for this story?**
+    * This is personal experience. Specifically, I had to do [almost exactly this dance][RuffleAsync] in order to get fetch to work in Ruffle.
+    * I have omitted a detail from this story: in Ruffle, we use a GC library (`gc_arena`) that imposes a special lifetime on all GC references. This is how the GC library upholds it's memory safety invariants, but it's also what forces us to pass around contexts, and once you have that, it's natural to start putting even non-GC data into it. It also means we can't hold anything from the GC in the Future as we cannot derive it's `Collect` trait on an anonymous type.
+* **Why did you choose Niklaus to tell this story?**
```
rylev, line 82: I'm surprised by the choice of Niklaus since to me Niklaus is supposed to represent someone with little programming experience. The protagonist of this story seems to be fairly well versed in programming. They seem more like an Alan - familiar with JavaScript and gaining experience in Rust but not an expert.

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+## The story
+
+Niklaus heard about a project to reimplement a deprecated browser plugin using Rust and WASM. This old technology had the ability to load resources over HTTP; so it makes sense to try and implement that functionality using the Fetch API. He looks up the documentation of `web_sys` and realizes that he needs to take a thing called a `Future` out of the call to `fetch`, do some things to it in Rust, and then send the future off to `spawn_local` as such:
+
+```use wasm_bindgen_futures::spawn_local;
+use web_sys::window;
+
+fn load_image(src: &url) {
+    spawn_local(async move {
+        let url = src.to_string();
```
nikomatsakis, line 18: This probably can't be part of the `spawn_local` call, or Niklaus would have the same problem, right? Maybe move it out?

src/vision/status_quo/niklaus_lost_his_context.md
nikomatsakis, line 45: ...might be nice to include an example of those error messages...but maybe too annoying to reproduce :)

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+
+        player.lock().unwrap().update(|context| {
+            let image = parse_png(data, context);
+            context.display_tree.add_child_node(inside_of, image);
+        });
+    });
+}
+```
+
+It works, but it causes problems. Async and contextual code can't really coexist; so async code always has to live at the periphery of the program. More complicated async behaviors, such as reading the image in chunks and progressively decoding it, requires a delicate dance of await followed by update and so on. Niklaus tells Alan about this and he says it has something to do with colored functions, something he isn't quite familiar about.
```
nikomatsakis, line 68:
```suggestion
It works, but it causes problems. Async and contextual code can't really coexist; so async code always has to live at the periphery of the program. More complicated async behaviors, such as reading the image in chunks and progressively decoding it, requires a delicate dance of await followed by update and so on. Niklaus tells Alan about this and Alan says it has something to do with colored functions, something he isn't quite familiar about.
```

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+
+## 🤔 Frequently Asked Questions
+
+* **What are the morals of the story?**
+    * Borrowing means that Rust has three function colors, not two:
+      * Sync, which can both own and borrow it's parameters
+      * Owned Async, which can only own parameters
+      * Borrowed Async, which can both own and borrow parameters, but gains a lifetime
+    * Non-`'static` Futures are of limited use as lifetimes are tied to the sync stack. I think you can use `block_on` with them but that's about it.
+    * Borrowed Async functions would be useful if there was a way to have a future whose borrows were scoped to the moment they are `poll`ed.
```
rylev, line 78: These seem to be the real crux of the issue. This makes it clearer to me that seeing the error messages would really drive this point home. Perhaps it would be good to show more of the struggle in the time between getting the initial error message and figuring out a solution.

rylev (changes requested):
This definitely has potential to be a really nice story about the struggles on making borrows work in the async world, but I don't think it's quite there yet. More detail about the compiler errors encountered and the struggle to get those errors fixed would be really great.

src/vision/status_quo/niklaus_lost_his_context.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+Niklaus adds calls to `load_image` where appropriate, sees the message pop up in the console, and figures it's time to now actually do something to that loaded image. At this point, it's important to note that many parts of this project pass around "contexts" - bundles of borrowed state that various parts of the reimplementation are free to alter. (This is necessary because this plugin had multiple scripting runtimes in it.) In synchronous code, this was perfectly fine. However, when we start integrating the function into the player...
+
+```use crate::{Index, Context}
+use wasm_bindgen_futures::spawn_local;
+use web_sys::window;
+use crate::parse_png;
+
+fn load_image(src: &url, inside_of: Index<Node>, context: &mut Context<'_>) {
```
nikomatsakis, line 34: I'd like to know a bit more about where this `context` comes from. =) I guess it is under Alan's control. I'm trying to figure out if this overall plugin is 'conceptually single threaded" or what -- could `fn load_image` be an `async fn`? Presumably not.

It sounds like what Alan probably wants is this `context` to be "ambiently available" to the futures, and to rely on the fact that there should be only one running at a time so that it can be `&mut`? Is it kind of way to setup context and for things to communicate?

nikomatsakis (commented):
This was fun to read! Some clarifying questions to help me understand what's going on.

src/vision/status_quo/niklaus_lost_his_context.md
kmeisthax: I imagine Niklaus knows about them in the sense that...

1. `fetch` lets him call out to JavaScript's `window.fetch`
2. `spawn_local` takes a Rust `Future` and somehow makes it "go"

He presumably does not (yet) understand that `spawn_local` needs to own the futures it runs, or why the lifetimes work out this way. Or why the application he's PRing to is architectured in such a way that makes async difficult.

Reading your other comment I think Alan might have been the better choice and I may rewrite this SQ accordingly.

src/vision/status_quo/niklaus_lost_his_context.md
kmeisthax: You are correct, it is multithreaded - half because we're in WASM and WASM multithreading is a mess right now, and half because the actual thing we're reimplementing also lived in a browser mainthread.

You're right about the overall intent, too. In fact, I probably should rename it to `World` to avoid confusion with executor contexts.

kmeisthax:
I've rewritten a good chunk of this story according to the feedback from @nikomatsakis and @rylev.

nikomatsakis (approved):
This is so much better, thanks! It was really clear and easy to read now.

src/vision/status_quo/alan_lost_the_world.md
```diff
@@ -0,0 +1,161 @@
...
+        world.nodes.push(image.into());
+    });
+}
+```
+
+It works, well enough that Alan is able to finish his changes and PR them into the project. However, Alan wonders if this could be syntactically cleaner, somehow. Right now, async and update code have to be separated - if we need to do something with a `World`, then `await` something else, that requires jumping in and out of this `update` thing. It's a good thing that we only really *have* to be async in these loaders, but it's also a shame that we practically *can't* mix `async` code and `World`s.
+
+## 🤔 Frequently Asked Questions
+
+* **What are the morals of the story?**
```
nikomatsakis, line 136: Maybe something like...

"Async functions capture all of their parameters for the entire duration of the function. This allows them to hold borrows of those parameters across await points. However, when the parameter represents the "global environment' in which the future runs, such as the `World` in this story, it may be useful for that parameter not to be captured by the future but rather supplied anew after each await point.

nikomatsakis (changes requested):
One thought

rylev (commented):
Awesome stuff! It's really getting there, but it needs a few more adjustments before it's ready.

src/vision/status_quo/alan_lost_the_world.md
```diff
@@ -0,0 +1,161 @@
...
+    * Borrowing means that Rust has three function colors, not two:
```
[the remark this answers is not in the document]
kmeisthax: This is in reference to the blog post "[What color is your function](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/)", which explains the additional complexity caused by adding explicit facilities for asynchronous code into a language. I'm altering the status-quo to explain this better.

src/vision/status_quo/alan_lost_the_world.md
```diff
@@ -0,0 +1,161 @@
...
+
+    num_times_called
+}
+```
+
+The function passed to `sort_by` has to borrow something not passed into it, namely the `num_times_called` variable. Hence, it has the lifetime of that borrow, not the whole program, because it can't be called anytime - only when `num_times_called` is a valid thing to read or write.
+
+Async functions, it turns out, *close over all their parameters*! They *have to*, because the futures they return don't actually *have* function parameters that the user controls. So they have to copy everything they get passed into, and if that thing is borrowed, then the entire future is borrowed.
+
+Barbara suggests changing all of the async function's parameters to be owned types. Alan asks Grace, who architected this project. Grace recommends holding a reference to the `Plugin` that owns the `World`, and then borrowing it whenver you need the `World`. That ultimately looks like the following:
```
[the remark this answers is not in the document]
kmeisthax: There was, I left an old version of that paragraph around which I have since removed.

src/vision/status_quo/alan_lost_the_world.md
```diff
@@ -89,7 +94,7 @@ where
 
 So, `spawn_local` only works with futures that return nothing - so far, so good - and are `'static`. Uh-oh. What does that last bit mean? Alan asks Barbara, who responds that it's the lifetime of the whole program. Yeah, but... the async function is part of the program, no? Why wouldn't it have the `'static` lifetime? Does that mean all functions that borrow values aren't `'static`, or just the async ones?
 
-Barbara explains that when you borrow a value in a function, the function doesn't gain the lifetime of that borrow. Instead, the borrow comes with it's own lifetime, separate from the function's. The only time a function can have a non-`'static` lifetime is if the function *closes over* some other variable, like so:
+Barbara explains that when you borrow a value in a closure, the closure doesn't gain the lifetime of that borrow. Instead, the borrow comes with it's own lifetime, separate from the closure's. The only time a closure can have a non-`'static` lifetime is if one or more of it's borrows is *not* provided by it's caller, like so:
```
nikomatsakis, line 97:
```suggestion
Barbara explains that when you borrow a value in a closure, the closure doesn't gain the lifetime of that borrow. Instead, the borrow comes with it's own lifetime, separate from the closure's. The only time a closure can have a non-`'static` lifetime is if one or more of its borrows is *not* provided by it's caller, like so:
```

nikomatsakis (changes requested):

nikomatsakis (commented):
A few minor nits from the last read... will read more in a sec...

nikomatsakis (approved):

nikomatsakis:
I think this is ready, going to merge!

Issue #46, opened 2021-03-18
learning async rust for the first time


nikomatsakis:
* Character: [Niklaus](https://rust-lang.github.io/wg-async-foundations/vision/characters/niklaus.html), but could be anyone but [Barbara](https://rust-lang.github.io/wg-async-foundations/vision/characters/barbara.html) I think
* Brief summary: Niklaus is going to learn Async Rust, but it's a lot. He has to [learn a lot of things at once](https://twitter.com/richardsabow/status/1345815109201842178). He is [afraid to pick the wrong runtime](https://twitter.com/EchoRior/status/1359964691305496579), especially as it is [tied to the kinds of libraries he may get to use later](https://zulip-archive.rust-lang.org/187312wgasyncfoundations/59166portabilityandruntimes.html#228879553). Perhaps he loves [Warp](https://crates.io/crates/warp) but wants to use it with [Bastion](https://bastion.rs/), because he digs Erlang.
* Key points or morals (if known):
    * People are very often learning all of Rust and Async Rust together and it can be overwhelming
    * Too many choices at every point in the step makes it hard
    * Picking a runtime is often stressful for a lot of folks, they are surprised they have to do it at all
    * Not sure what else! More I bet.
* Sources:
    * [Notes from twitter thread](https://rust-lang.github.io/wg-async-foundations/conversations/2021-02-12-Twitter-Thread.html)
    * [Relevant Zulip discussion](

nikomatsakis:
This [comment from reddit](https://www.reddit.com/r/rust/comments/m7u6gg/building_a_shared_vision_for_async_rust/grft841/) seems like related data:

> $0.02
>
> I've been both a c++ developer and a developer in a gc language, and I've previously gotten as far as writing a sudoku solver in rust, but no further. It had some enums, but generally avoided most issues of borrowing with bitpacking into a fixed sized object and copying.
>
> I check in on rust every now and again and would like to write more in it, but eventually most of my problems need async solutions.
>
> However, I bounce off rust every time, usually at a point where I'm ok with basic syntax. At that point I've either looked for something i can get my teeth into to that forces me to confront lifetimes, or I started looking for async docs to do something useful and saw that they were at the time largely left as to-do in the rust book (iirc).
>
> The impression that async isn't ready knocks the motivation out of me. I'm looking for a solution, not a work in progress, and I don't have a pressing problem that requires me to solve it in rust.
>
[...]

PR #70, opened 2021-03-19
status quo: Alan started trusting the Rust compiler

MidasLamb:
[Rendered](https://github.com/MidasLamb/wg-async-foundations/blob/patch-1/src/vision/status_quo/alan_started_trusting_the_rust_compiler.md)

PR #94, opened 2021-03-26
Add a status quo story about Barbara's first async playground

eminence:
This is a status quo story about Barbara and her first experiences with rust async.  It was developed mostly from my personal experiences, plus notes from a group discussion with some folks on a zoom "writing session" organized by Niko.

Of the current open issues, this one most closely matches #46.

This new story here is pretty similar to [Barbara makes their first foray into async](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_makes_their_first_steps_into_async.html).   Both this story and that story start with Barbara in the same place, but they evolve a bit differently and focus in on different things.  My first inclination is to keep both stories, but we can certainly discuss if they are worth merging.   We might want to adjust the titles of these a bit to make sure they are sufficiently distinct.  

[Rendered](https://github.com/rust-lang/wg-async-foundations/blob/60c95937df919566404b15d1d8f8ee9048a4f560/src/vision/status_quo/barbara_plays_with_async.md)

### DIFF
--- a/src/vision/status_quo/barbara_plays_with_async.md
+++ b/src/vision/status_quo/barbara_plays_with_async.md
@@ -23,7 +23,7 @@ project, she starts with a "playground" where she can try to write some simple
 async rust code to see how it feels and how it compares to how async code feels
 in other languages she knows (like C# and JavaScript).
 
-She starts by open a blank project in VSCode with [rust-analyzer].  Because she's
+She starts by opening a blank project in VSCode with [rust-analyzer].  Because she's
 been following the overall state of rust async, she knows that she needs a runtime,
 and quickly decides to use tokio, because she knows its quite popular and well documented.
 
@@ -227,9 +227,9 @@ let mut futures: Vec<Box<dyn Future<Output=()>>> = Vec::new();
 She then notices that her IDE (VSCode + rust-analyzer) has a new error on
 each call to push.  The code assist on each error says `Store this in the heap
 by calling 'Box::new'`.  She is exactly what she wants, and it happy that
-rust-analzyer perfectly handled this case.
+rust-analyzer perfectly handled this case.
 
-Now each future is boxed up, ut there is one final error yet,
+Now each future is boxed up, but there is one final error still,
 this time on the call to `join_all(futures).await`:
 
 ```
@@ -279,7 +279,7 @@ error[E0658]: async closures are unstable
    = help: to use an async block, remove the `||`: `async {`
 ```
 
-Barbara knows that async are stable and using this nightly feature isn't what
+Barbara knows that async is stable and using this nightly feature isn't what
 she wants.  So the tries the suggestion made by the compiler and removes the `||` bars:
 
 ```rust
--- a/src/vision/status_quo/barbara_plays_with_async.md
+++ b/src/vision/status_quo/barbara_plays_with_async.md
@@ -7,7 +7,7 @@ This is a draft "status quo" story submitted as part of the brainstorming period
 
 If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
 
-# The story
+## The story
 
 [Barbara] has been following async rust for a long time, in eager anticipation
 of writing some project using async.  The last time she tried to do anything
@@ -328,7 +328,7 @@ futures.push(Box::pin(async move {
 It works!  Satisfied but still thinking about async rust, Barbara takes a break
 to eat a cookie.
 
-# 🤔 Frequently Asked Questions
+## 🤔 Frequently Asked Questions
 
 *Here are some standard FAQ to get you started. Feel free to add more!*
 
--- a/src/vision/status_quo/barbara_plays_with_async.md
+++ b/src/vision/status_quo/barbara_plays_with_async.md
@@ -243,8 +243,22 @@ Barbara has been around rust for long enough to know that there is a `Box::pin`
 API, but she doesn't really understand what it does, nor does she have a good
 intuition about what this API is for.  But she is accustomed to just trying
 things in rust to see if they work.  And indeed, after changing `Box::new` to
-`Box::pin` (and adjusting the type of the Vec), the code compiles and runs
-successfully.
+`Box::pin`:
+
+```rust
+futures.push(Box::pin(delay_future.then(|_| {
+    println!("Done!");
+    std::future::ready(())
+})));
+```
+
+and adjusting the type of the Vec:
+
+```rust
+let mut futures: Vec<Pin<Box<dyn Future<Output=()>>>> = Vec::new();
+```
+
+the code compiles and runs successfully.
 
 But even though the run is working correctly, she wishes she had a better idea
 why pinning is necessary here and feels a little uneasy having to use something
--- a/src/vision/status_quo/barbara_plays_with_async.md
+++ b/src/vision/status_quo/barbara_plays_with_async.md
@@ -1,0 +1,387 @@
+# 😱 Status quo stories: Barbara plays with async
+
+
+## 🚧 Warning: Draft status 🚧
+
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+* What are other related stories?
+    * [Barbara makes their first steps in async] is Barbara in a slightly different universe.
+    * [Alan started trusting the rust compiler][trusts the rust compiler] is a similar story about a different character.
+
+[status quo stories]: ./status_quo.md
+[Barbara]: ../characters/barbara.md
+[htvsq]: ../how_to_vision/status_quo.md
+[cannot be wrong]: ../how_to_vision/comment.md#comment-to-understand-or-improve-not-to-negate-or-dissuade
+[trusts the rust compiler]: ./alan_started_trusting_the_rust_compiler_but_then_async.md
+[Barbara makes their first steps in async]: ./barbara_makes_their_first_steps_into_async.md
```
eminence, line 379: Note: this is a dead link until #70 is merged

src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+This was not a pleasant experience for Barbara.
+
+After watching the development of rust async/await (by following
+discussions on [/r/rust][reddit] and the [internals] forums), she wants
+to start to play around with writing async code.  Before starting on any real
+project, she starts with a "playground" where she can try to write some simple
+async rust code to see how it feels and how it compares to how async code feels
+in other languages she knows (like C# and JavaScript).
+
+She starts by open a blank project in VSCode with [rust-analyzer].  Because she's
```
rylev, line 26:
```suggestion
She starts by opening a blank project in VSCode with [rust-analyzer].  Because she's
```

src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+```rust
+let mut futures: Vec<Box<dyn Future<Output=()>>> = Vec::new();
+```
+
+She then notices that her IDE (VSCode + rust-analyzer) has a new error on
+each call to push.  The code assist on each error says `Store this in the heap
+by calling 'Box::new'`.  She is exactly what she wants, and it happy that
+rust-analzyer perfectly handled this case.
+
+Now each future is boxed up, ut there is one final error yet,
```
rylev, line 232:
```suggestion
Now each future is boxed up, but there is one final error still,
```

src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+   |
+14 |         futures.push(Box::pin(async || {
+   |                               ^^^^^
+   |
+   = note: see issue #62290 <https://github.com/rust-lang/rust/issues/62290> for more information
+   = help: add `#![feature(async_closure)]` to the crate attributes to enable
+   = help: to use an async block, remove the `||`: `async {`
+```
+
+Barbara knows that async are stable and using this nightly feature isn't what
```
rylev, line 282:
```suggestion
Barbara knows that async isn't stable and using this nightly feature isn't what
```

src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+Barbara has been around rust for long enough to know that there is a `Box::pin`
+API, but she doesn't really understand what it does, nor does she have a good
+intuition about what this API is for.  But she is accustomed to just trying
+things in rust to see if they work.  And indeed, after changing `Box::new` to
+`Box::pin` (and adjusting the type of the Vec), the code compiles and runs
+successfully.
+
+But even though the run is working correctly, she wishes she had a better idea
```
rylev, line 249: I like this point. Barbara got the code working, but has an uneasy feeling about it, since she doesn't fully understand *why* it's needed.

rylev (commented):
This looks really great! I only have some small comments.

src/vision/status_quo/barbara_plays_with_async.md
eminence: Actually, what I wanted to say is:  "Barbara knows that async *is* stable".  Do I have that right?

eminence:
Thanks for the typo fixes @rylev !

Skimming over the notes made on the hackmd document, here are some things that we talked about adding to the story, but didn't actually make it into this version:

* Using the IDE to discover API (by invoking tab completion and scrolling the through the results) doesn't really work here (though this story talks about something similar in the FAQ section)
* Didn't really show any "Errors from hell" 
* Didn't describe that Barbara had to go some google searches, and how she found the rust async book (that probably would have been a sensible thing for Alan to do, who was the original character for this story.  But since we changed from Alan to Barbara, perhaps we don't need to include this part)

src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+error[E0277]: `()` is not a future
+  --> src\main.rs:13:71
+   |
+13 |         futures.push(tokio::time::sleep(Duration::from_millis(delay)).then(|_| {
+   |                                                                       ^^^^ `()` is not a future
+   |
+   = help: the trait `futures::Future` is not implemented for `()`
+```
+
+Even though the error is pointing at the `then` function, Barbara pretty quickly
```
nikomatsakis, line 98: Hard to say how it would be if I were editing the code, but fwiw I stared at this quite a long time before reading on to understand what was happening. I could see this bug being quite confusing. =)

src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+She first adds explicit type annotations to the Vec:
+
+```rust
+let mut futures: Vec<Box<dyn Future<Output=()>>> = Vec::new();
+```
+
+She then notices that her IDE (VSCode + rust-analyzer) has a new error on
+each call to push.  The code assist on each error says `Store this in the heap
+by calling 'Box::new'`.  She is exactly what she wants, and it happy that
+rust-analzyer perfectly handled this case.
```
nikomatsakis, line 230:
```suggestion
rust-analyzer perfectly handled this case.
```

src/vision/status_quo/barbara_plays_with_async.md
```diff
@@ -0,0 +1,387 @@
...
+   |
+34 |     futures::future::join_all(futures).await;
+```
+
+Barbara has been around rust for long enough to know that there is a `Box::pin`
+API, but she doesn't really understand what it does, nor does she have a good
+intuition about what this API is for.  But she is accustomed to just trying
+things in rust to see if they work.  And indeed, after changing `Box::new` to
+`Box::pin` (and adjusting the type of the Vec), the code compiles and runs
+successfully.
```
nikomatsakis, line 247: Code sample?

nikomatsakis (approved):
This is great! Left a few small comments.

src/vision/status_quo/barbara_plays_with_async.md
eminence: I pushed a commit that shows how each part changed.  In order to avoid repeating basically the same block of code over-and-over, I just showed the changed parts.  But if you think it'll read better to show the whole thing, let me know.

src/vision/status_quo/barbara_plays_with_async.md
eminence: This is a great point.  For me personally, I was able to draw upon experiences from years ago trying to use futures in a pre-async world, where you can get some pretty gnarly type errors.   But I don't know if we want to give Barbara those same experiences, or if we want to have her struggle a bit more with this particular error.

Also, when reviewing this code again I realized that the `then` method is only available as part of the `futures` crate, however, we haven't yet introduced this crate as this point in the story.  I kinda feel like this part of the story should be rewritten to fix this inconsistency.  What do you think?

nikomatsakis: Yeah, it seems ok to have her know about futures already and go on the lookout for combinators.

eminence: 👍 ok gonna mark this as 'resolved' for now

doc-jones (commented):
Are the template subheadings optional?  I thought we were all using those.

PR #97, opened 2021-03-26
Alan tries using a socket Sink

cryptoquick:
This is a personal story, but I've done my best to switch things up so it's less personal, while also doing my best to keep the details more grounded in the specifics. I also tried to make it follow a similar format to the others.

Hopefully the "drama" portion isn't too distracting, but I think it's important to explain what happened, because we need to acknowledge the issue and be honest about it.

Of course, I'm open to suggestions. I also will admit, I had trouble working on the second example. I'm still not sure what the best solution is with that approach.


cryptoquick:
Also, I just realized, I was supposed to use an existing persona. I'll update this, give me a few. I originally chose the name Carrie just because of... Well, personal reasons.

cryptoquick:
I've updated this, and I'd like feedback on either echo examples, as well as suggestions for the FAQ section.

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+
+## The story
+
+Alan is working on a project that uses an alternative popular async framework. He finds the new async framework less familiar, and as he reads through their GitHub issues, he's saddened by the project maintainer's unwelcoming attitude towards their community. He asks his teammates if they can switch to the one he's more familiar with and feels more welcome in, but the teammates are understandably resistant to changing something that important just for one feature. So, he works hard to find workarounds to accomplish his team's goals.
+
+One of the goals is to switch from a WebSocket implementation using raw TCP sockets to one managed behind an HTTP server library, so both HTTP and WebSocket commands can be forwarded to a transport-agnostic RPC server. He finds an HTTP server that's similar to one he's used to using with the other async framework, and a WebSocket middleware library that goes with it.
+
+However, as he's working, Alan encounters a situation where the socket needs to be written to within an async thread, and the traits just aren't working. He wants to do split the stream into a sender and receiver:
```
ibraheemdev, line 15:
```suggestion
However, as he's working, Alan encounters a situation where the socket needs to be written to within an async thread, and the traits just aren't working. He wants to split the stream into a sender and receiver:
```

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+    debug!("Received new WS RPC message: {:?}", msg);
+
+    // Echo the request:
+    match ws_sender.lock().await.send_string(msg).await {
+        Ok(_) => {
+            info!("New WS data sent.");
+        }
+        Err(_) => {
+            warn!("WS connection closed.");
+        }
```
ibraheemdev, line 36:
```suggestion
        Ok(_) => info!("New WS data sent."),
        Err(_) => warn!("WS connection closed."),
```

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+    };
+}
+```
+
+This is necessary because both sides of the stream are stateful Iterables, if they're kept as one, the lock on the receiver couldn't be released within the loop. He's seen this pattern used in other projects in the Rust community, but is frustrated to find that the Sink trait wasn't implemented in the WebSockets middleware library he's using.
```
ibraheemdev, line 41:
```suggestion
This is necessary because both sides of the stream are stateful iterators, and if they're kept as one, the lock on the receiver can't be released within the loop. He's seen this pattern used in other projects in the Rust community, but is frustrated to find that the `Sink` trait wasn't implemented in the WebSockets middleware library he's using.
```

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+
+async_std::task::spawn(async move {
+    while let Some(msg) = ws_receiver.next().await {
+        match poller_ws_stream.lock().await.send_string(msg).await {
+            Ok(msg) => {
+                info!("New WS data sent. {:?}", msg);
+            }
+            Err(msg) => {
+                warn!("WS connection closed. {:?}", msg);
+            }
```
ibraheemdev, line 65:
```suggestion
            Ok(msg) => info!("New WS data sent. {:?}", msg),
            Err(msg) => warn!("WS connection closed. {:?}", msg),
```

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+    }
+});
+
+while let Some(msg) = ws_stream.lock().await.next().await {
+    // Echo the request?
+    ws_sender.send(msg);
+}
+```
+
+Alan wonders if he's thinking about it wrong, but the solution isn't as obvious as his earlier Sink approach. Looking around, he realizes a solution to his problems already exists-- as others have been in his shoes before-- within two other nearly-identical pull requests, but they were both closed by the project maintainers. He tries opening a third one with the same code, pointing to an example where it was actually found to be useful. To his joy, his original approach works with the code in the closed pull requests in his local copy! Alan's branch is able to compile for the first time.
```
ibraheemdev, line 76:
```suggestion
Alan wonders if he's thinking about it wrong, but the solution isn't as obvious as his earlier `Sink` approach. Looking around, he realizes a solution to his problems already exists-- as others have been in his shoes before-- within two other nearly-identical pull requests, but they were both closed by the project maintainers. He tries opening a third one with the same code, pointing to an example where it was actually found to be useful. To his joy, his original approach works with the code in the closed pull requests in his local copy! Alan's branch is able to compile for the first time.
```

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+
+However, almost immediately, his request was also shut down, and the maintainer suggested he try the complex, unobvious, and unspecific solution that he had already tried and just couldn't get it to work.
```
ibraheemdev, line 78:
```suggestion
However, his pull request gets shut down instantaneously, and the maintainer suggests he try the complex, unobvious, and unspecific solution that he had already tried and couldn't get it to work.
```

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+
+As a result of his frustration, Alan calls out one of the developers of the project on social media, who he believed had frequently shot down the idea of using Sink traits on projects others needed to use them on. This is not well-received, the maintainer reacts with a condescending and dismissive attitude, and he later finds out he was blocked. A co-maintainer responds to the thread, defending and supporting the other maintainer's actions, and he's told to "get over it". He's given a link to a blog post with a piece [lamenting the popularity of Sink in the Rust ecosystem](https://blog.yoshuawuyts.com/rust-streams/#why-we-do-not-talk-about-the-sink-trait), and how it's complex and bad and not worth it, but the piece also unhelpfully makes no effort to provide examples of how the better alternatives would be used to replace uses of Sink.
```
ibraheemdev, line 80:
```suggestion
As a result of his frustration, Alan calls out one of the developers of the project on social media, who he believed had frequently shot down the idea of using `Sink` traits. This is not well-received, the maintainer reacts with a condescending and dismissive attitude, and he later finds out he was blocked. A co-maintainer responds to the thread, defending and supporting the other maintainer's actions, and he's told to "get over it". He's given a link to a blog post with a piece [lamenting the popularity of Sink in the Rust ecosystem](https://blog.yoshuawuyts.com/rust-streams/#why-we-do-not-talk-about-the-sink-trait), and how it's complex,  bad and not worth it. The piece unhelpfully makes no effort to provide examples of how the better alternatives would be used to replace uses of `Sink`.
```

src/vision/status_quo/carrie_tries_a_socket_sink.md
```diff
@@ -0,0 +1,113 @@
...
+
+Because of this heated exchange, Alan grows concerned for his own career, what these well-known community members might think or say about his to others, and his confidence in the community surrounding this language that he really enjoys using is somewhat shaken.
+
+Despite this, Alan takes a walk, gathers his determination, and commits to maintaining his fork with the changes from the other pull requests that were shut down, and publihes his version to Crates.io, vowing to be more welcoming to "misfit" pull requests like the one he needed.
```
ibraheemdev, line 84:
```suggestion
Despite this, Alan takes a walk, gathers his determination, and commits to maintaining his fork with the changes from the other pull requests that were shut down, and publishes his version to crates.io, vowing to be more welcoming to "misfit" pull requests like the one he needed.
```

ibraheemdev (changes requested):
Just a few minor nits. Overall, I think this is really well done, yet sad at the same time because I've had a similar experience.

src/vision/status_quo/carrie_tries_a_socket_sink.md
ibraheemdev: I'm not sure if you meant to name the specific blog post here?

src/vision/status_quo/carrie_tries_a_socket_sink.md
cryptoquick: I think instantaneously is the wrong phrase here, but the other suggestions are great!

src/vision/status_quo/carrie_tries_a_socket_sink.md
cryptoquick: Do you think I should put it down below?

cryptoquick:
@ibraheemdev Really good suggestions, thank you! And I appreciate the support.

jbr:
The other side of this story is relevant to this wg.

The short version is that because the futures crate is treated as if it were stdlib, *not* implementing traits from that library is seen as breaking from convention. If the futures crate were to deprecate Sink (or alternatively if less likely for there to be action on moving Sink into std and therefore async-std), this would have been easier to navigate as a framework author. The awkward implementation of the code in question was explicitly intended to take a Sink based api and hide the Sink trait. The alternative was to fork async-tungstenite and make the changes there.

Summary: The futures crate is treated like std and Sink is seen as comparably std-track as Stream and Future and there is nothing to reference as a library author that indicates whether this is or is not the case

jbr:
@cryptoquick reading the code in this story: The connection is Clone, so there's no reason to re-wrap it with an Arc Mutex. You can just clone it as many times as you want and do whatever you'd do with a stream or a sink.

For other readers: I'm sorry this part spilled over here. The relevant bit is that "implement Sink" was the feature request, not "implement Clone"

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,106 @@
...
+
+## The story
+
+Alan is working on a project that uses an alternative popular async framework. He finds the new async framework less familiar, and as he reads through their GitHub issues, he's saddened by the project maintainer's unwelcoming attitude towards their community. He asks his teammates if they can switch to the one he's more familiar with and feels more welcome in, but the teammates are understandably resistant to changing something that important just for one feature. So, he works hard to find workarounds to accomplish his team's goals.
+
+One of the goals is to switch from a WebSocket implementation using raw TCP sockets to one managed behind an HTTP server library, so both HTTP and WebSocket commands can be forwarded to a transport-agnostic RPC server. He finds an HTTP server that's similar to one he's used to using with the other async framework, and a WebSocket middleware library that goes with it.
+
+However, as he's working, Alan encounters a situation where the socket needs to be written to within an async thread, and the traits just aren't working. He wants to split the stream into a sender and receiver:
```
nikomatsakis, line 15: I'm having a hard time understanding this, actually, there seems to be some context I am missing. Poking at the rustdoc, I see the [`split`](https://docs.rs/futures/0.3.13/futures/stream/trait.StreamExt.html#method.split) method, that works on a `Stream + Sink`. Is the reason that this code doesn't work because `ws_stream` does not implement `Sink`? That could be made clearer -- for example, you could show the error message he gets.

nikomatsakis (commented):
Ack! I ran out of time. I'm going to read this over. I don't have a problem with including some amount of drama, and I would even say it's good to include some element of that, since I think it's something of an Elephant in the Room so much of the time, but I also think we should be careful how we phrase it. I haven't really read that part of the text closely enough to make any suggestions, I will attempt to do so later.

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,106 @@
...
+
+Despite this, Alan takes a walk, gathers his determination, and commits to maintaining his fork with the changes from the other pull requests that were shut down, and publishes his version to crates.io, vowing to be more welcoming to "misfit" pull requests like the one he needed.
+
+A few weeks later, Alan's work at his project at work is merged with his new forked crate. It's a big deal, his first professional open source contribution to a Rust project, but he still doesn't feel like he has a sense of closure with the community. Meanwhile, his friends say they want to try Rust, but they're worried about its async execution issues, and he doesn't know what else to say, other than to offer a sense of understanding. Maybe the situation will get better someday, he hopes.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+* **What are the morals of the story?**
```
eminence, line 84: This whole story is obviously about a real and specific event, which is great (we like specifics in these kind of stories).

But as I'm coming to the end of the story and thinking about morals, I'm also trying to draw generalizations too.  Based on these comments:

> but is frustrated to find that the Sink trait wasn't implemented in the WebSockets middleware library he's using.

> lamenting the popularity of Sink in the Rust ecosystem, and how it's complex, bad and not worth it.

I'm wondering if there is also something to say here about the difficulties/friction in trying to get multiple async libraries to interop, *especially* when they have conflicting design philosophies.

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,106 @@
...
+
+while let Some(msg) = ws_stream.lock().await.next().await {
+    // Echo the request?
+    ws_sender.send(msg);
+}
+```
+
+Alan wonders if he's thinking about it wrong, but the solution isn't as obvious as his earlier `Sink` approach. Looking around, he realizes a solution to his problems already exists-- as others have been in his shoes before-- within two other nearly-identical pull requests, but they were both closed by the project maintainers. He tries opening a third one with the same code, pointing to an example where it was actually found to be useful. To his joy, his original approach works with the code in the closed pull requests in his local copy! Alan's branch is able to compile for the first time.
+
+However, almost immediately, his request was shut down like the others, and the maintainer suggests he try the complex, unobvious, and unspecific solution that he had already tried and just couldn't get it to work.
```
eminence, line 70: If it's not too long, is it possible to have a snippet of code that shows the code that Alan finds complex/unobvious/unspecific?

nikomatsakis: That would indeed be very helpful

nikomatsakis (commented):
Hi @cryptoquick! I left various comments with suggested alterations to the wording. My goal is to keep the sense of your original story, but in general to avoid ascribing intentions or motivations to other people. I think a good goal should be to keep your description of other's actions short and factual -- particularly negative ones -- and instead ascribe feelings and interpretations to Alan.

For example, rather than saying "the maintainer sent a dismissive reply", I wrote "Alan felt dismissed".

I have to say that i would *also* really like to see the technical aspects of this story increased. I've had some confusion over the role of `Sink` myself, and I think it'd be helpful to dig into. I left some comments to that effect already, and I added a few more here. Maybe it'd be nice to just open a different story about `Sink` though.

src/vision/status_quo/alan_tries_a_socket_sink.md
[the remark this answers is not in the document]
cryptoquick: Great way to explain it!

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,106 @@
...
+    match ws_sender.lock().await.send_string(msg).await {
+        Ok(_) => info!("New WS data sent."),
+        Err(_) => warn!("WS connection closed."),
+    };
+}
+```
+
+This is necessary because both sides of the stream are stateful iterators, and if they're kept as one, the lock on the receiver can't be released within the loop. He's seen this pattern used in other projects in the Rust community, but is frustrated to find that the `Sink` trait wasn't implemented in the WebSockets middleware library he's using.
+
+Alan also tries creating a sort of poller worker thread using an intermediary messaging channel, but he has trouble reasoning about the code and wasn't able to get it to compile:
```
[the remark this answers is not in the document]
cryptoquick: Yes, it's suggested here:
https://github.com/http-rs/tide-websockets/issues/15#issuecomment-797090892

That said, I also tried implementing it that way beforehand.

src/vision/status_quo/alan_tries_a_socket_sink.md
cryptoquick: That was what the second example was for, I'm not sure if there's a better way to explain that?

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,106 @@
...
+
+    // Echo the request:
+    match ws_sender.lock().await.send_string(msg).await {
+        Ok(_) => info!("New WS data sent."),
+        Err(_) => warn!("WS connection closed."),
+    };
+}
+```
+
+This is necessary because both sides of the stream are stateful iterators, and if they're kept as one, the lock on the receiver can't be released within the loop. He's seen this pattern used in other projects in the Rust community, but is frustrated to find that the `Sink` trait wasn't implemented in the WebSockets middleware library he's using.
```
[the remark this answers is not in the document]
cryptoquick: I'm not sure how better to explain it...

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,106 @@
...
+However, as he's working, Alan encounters a situation where the socket needs to be written to within an async thread, and the traits just aren't working. He wants to split the stream into a sender and receiver:
+
+```rust
+use futures::{SinkExt, StreamExt};
+use async_std::sync::{Arc, Mutex};
+use log::{debug, info, warn};
+
+/// In the connection handler:
+let (ws_sender, mut ws_receiver) = ws_stream.split();
+let ws_sender = Arc::new(Mutex::new(ws_sender));
```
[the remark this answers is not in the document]
cryptoquick: I added use of spawn and the await `rpc_call` example. So, it's less of an echo server and more related to my practical usage of it. Does that help?

src/vision/status_quo/alan_tries_a_socket_sink.md
cryptoquick: It's not really an error message so much as undesired locking behavior.

cryptoquick:
@eminence @nikomatsakis I've made a few more changes in response to your suggestions, can you take another look?

Also, I'm very grateful for your edits, @nikomatsakis, they were very helpful!

nikomatsakis (commented):
Ack! Apparently I never posted these comments, @cryptoquick, apologies. Let me give it a more thorough read.

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,115 @@
...
+
+A few weeks later, Alan's work at his project at work is merged with his new forked crate. It's a big deal, his first professional open source contribution to a Rust project! Still, he  doesn't feel like he has a sense of closure with the community. Meanwhile, his friends say they want to try Rust, but they're worried about its async execution issues, and he doesn't know what else to say, other than to offer a sense of understanding. Maybe the situation will get better someday, he hopes.
+
+## 🤔 Frequently Asked Questions
+
+
+* **What are the morals of the story?**
+    * There are often many sources of opinion in the community regarding futures and async, but these opinions aren't always backed up with examples of how it should be better accomplished. Sometimes we just find a thing that works and would prefer to stick with it, but others argue that some traits make implementations unnecessarily complex, and choose to leave it out. Disagreements like these in the ecosystem can be harmful to the reputation of the project and the participants.
```
nikomatsakis, line 89: I'm trying to decide what I think the morals of the story are. I would probably list:

* The `Sink` trait is complex but it solves a real problem, and the workarounds required to solve problems without it can be unsatisfactory. 
* Disagreement about core abstractions like `Sink` can make interoperability between runtimes more difficult; it also makes it harder for people to reproduce patterns they are used to from one runtime to another.
* It is all too easy for technical discussions like this to become heated; it's important for all participants to try and provide each other with the "benefit of the doubt".

nikomatsakis: I'm not sure if the final bullet is worth including, but it's definitely something I took away from this story.

src/vision/status_quo/alan_tries_a_socket_sink.md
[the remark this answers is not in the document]
nikomatsakis: It'd be great to include that citation in the story, I think, or just to explain why Alan thought this was a good thing to do.

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,115 @@
...
+use log::{debug, info, warn};
+
+/// In the connection handler:
+let (ws_sender, mut ws_receiver) = ws_stream.split();
+let ws_sender = Arc::new(Mutex::new(ws_sender));
+
+while let Some(msg) = ws_receiver.next().await {
+    debug!("Received new WS RPC message: {:?}", msg);
+
+    async_std::task::spawn(async move {
```
nikomatsakis, line 29: I think we need `let ws_sender = ws_sender.clone();`, right?

src/vision/status_quo/alan_tries_a_socket_sink.md
nikomatsakis: OK, I think maybe I'm starting to see. From reading the code snippet, what I get is this:

* He wants to have incoming messages and then spawn off threads that will do some RPC before forwarding the messages on to another place?

I think what's confusing to me is that I think of a stream as an "async iterator", and I don't quite get how it could be split. It seems like what you want is to "splice" yourself in the middle of the stream?

src/vision/status_quo/alan_tries_a_socket_sink.md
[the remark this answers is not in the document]
nikomatsakis: I think it does, see my comment above.

src/vision/status_quo/alan_tries_a_socket_sink.md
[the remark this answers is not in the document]
nikomatsakis: Like, I literally don't see a lock on `ws_receiver`, so that is still confusing to me. =) I guess the point is that `ws_receiver` can only produce one message at a time, but you want to introduce some parallelism after they are produced? (This is what I tried to note above, but elaborated)

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -0,0 +1,115 @@
...
+use log::{debug, info, warn};
+
+/// In the connection handler:
+let (ws_sender, mut ws_receiver) = channel::unbounded::<String>();
+let ws_receiver = Arc::new(ws_receiver);
+
+let ws_stream = Arc::new(Mutex::new(ws_stream));
+let poller_ws_stream = ws_stream.clone();
+
+async_std::task::spawn(async move {
```
nikomatsakis, line 56: I'm starting to see the idea here, but it is indeed still confusing to me. this snippet of code doesn't seem to be starting with the same "general inputs". Maybe it'd be helpful if we could give a bit of an architectural diagram?

nikomatsakis (changes requested):
OK, I gave more of a read. @cryptoquick I think I'm starting to get what's going on, but I'm wondering if you and I could arrange a time to dig into the technical details synchronously? I'd really like to understand the point you're making here more deeply but I just don't quite see it yet!

Maybe ping me on discord or zulip? Or send me an email at `niko@alum.mit.edu` with your availability?

src/vision/status_quo/alan_tries_a_socket_sink.md
cryptoquick: Yep! I think `async iterator` is a good way to think of a stream as long as it's coupled with the perspective that new items can be added to the stream, I don't think it has to be finite. It's kinda like a duplex unbounded MPSC. But streams vary a great deal on their implementation details, such as high water mark, and backpressure signalling, but that's a little beyond the scope of this.

That said, I'm coming at this from a NodeJS perspective, where stream use is common, and often preferred. A WebSocket is an ideal candidate for this, but it's important it can be both read and written to in exclusion.

src/vision/status_quo/alan_tries_a_socket_sink.md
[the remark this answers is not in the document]
cryptoquick: Yep, that's by design! `ws_receiver` in both cases is single-threaded, and thus does not need to be sent across multiple threads nor is it necessary to lock.

src/vision/status_quo/alan_tries_a_socket_sink.md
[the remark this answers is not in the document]
cryptoquick: It is! I think we're keeping the links in a separate section?

src/vision/status_quo/alan_tries_a_socket_sink.md
```diff
@@ -8,9 +8,9 @@ If you would like to expand on this story, or adjust the answers to the FAQ, fee
...
+Alan is working on a project that uses async-std. He has worked a bit with tokio in the past and is more familiar with that, but he is interested to learn something how things work in async-std.
 
-One of the goals is to switch from a WebSocket implementation using raw TCP sockets to one managed behind an HTTP server library, so both HTTP and WebSocket commands can be forwarded to a transport-agnostic RPC server. He finds an HTTP server that's similar to one he's used to using with the other async framework, and a WebSocket middleware library that goes with it.
+One of the goals is to switch from a WebSocket implementation using raw TCP sockets to one managed behind an HTTP server library, so both HTTP and WebSocket commands can be forwarded to a transport-agnostic RPC server. He finds the HTTP server tide and it seems fairly similar to warp, which he was using with tokio. He also finds the WebSocket middleware library *foobar* that goes with it.
```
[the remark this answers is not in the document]
cryptoquick: I did! :)

src/vision/status_quo/alan_tries_a_socket_sink.md
cryptoquick: That's an excellent way of putting it! I'll include your comments verbatim, I don't see anything I would change. Also, I really wish I hadn't assumed malice on the part of the maintainers, but I was just really frustrated by the lack of communication. I also wish I had been given the benefit of the doubt and not treated like some "troll".

src/vision/status_quo/alan_tries_a_socket_sink.md
cryptoquick: The only input that matters (and is common between both examples) is `ws_stream`. The implementation details differ, but I think they're roughly equivalent in practice.

cryptoquick:
Great feedback so far! I've pushed some more changes. I'm not sure if GitHub has some comment threads that are outdated, so please double-check.

src/vision/status_quo/alan_tries_a_socket_sink.md
nikomatsakis: Iterators don't have to be finite either, right?

Ultimately I think of iterators (and streams) as "producers" and functions (and sinks) as "consumers".  I was finding it very odd to think about something being `Stream + Sink`, but I guess it makes sense if you think of (say) a network port, or a bidirectional channel. I guess we are talking about a *socket* here, so that makes sense.

cryptoquick: Yep, I suppose you're right!

jbr: I'd love some clarification on the splice part of this. If there were a way to split the connection into a "sender" and a "receiver" wherein the receiver is Stream and the sender just has a send(&self, message: Message) associated function but which doesn't implement Sink, what would be missing?

As a separate question, if the same type that was the Stream also had the send function and was Clone so any clone of it could act as either side of that, what would be missing?

src/vision/status_quo/alan_tries_a_socket_sink.md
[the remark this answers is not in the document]
nikomatsakis: I usually include them inline if they are relevant to the story -- e.g., "He was browsing the issues and he saw XXX, which described a different approach..."

src/vision/status_quo/alan_tries_a_socket_sink.md
cryptoquick: So, if it's cloned, it wouldn't have to be locked by the sender thread, while also being read by the receiver loop?

cryptoquick: Also, I'm totally for improvements to `tide-websockets`, but is that not outside the scope of this PR?

jbr: It's worth figuring out if the Sink trait is necessary, since that's what this issue and the closed issue were about.

The latter situation I described where any clone could act as sender or receiver wouldn't require an improvement, it's the current state of the code. The sender and receiver are locked internally, so there can only be one clone sending at a time, and one clone polling the stream at a time, but they can be sending and receiving concomitantly

nikomatsakis: I think the splicing was a misunderstanding on my part. I eventually realized that what was happening was just receiving messages, doing some processing, and responding in kind.

nikomatsakis: Definitely changes to tide-websockets or whatever are not in scope for the story.

nikomatsakis: I don't think the *story itself* should describe whether this can be solved without `Sink`, but I would definitely love to explore possible resolutions to this story as "shiny futures", or possibly as FAQs. I think we can merge without that. I talked to @cryptoquick some -- one thing that I was explaining is that it should be possible for him to add a newtype wrapped around the `WebSocket` type that implements `futures::Sink` and `futures::Stream`, which would then allow him to write his original example without having to maintain a complete parallel fork of the crate. This seems better to me.

It does seem to me that *some* trait that means "can consume values of type `Item`" would be appropriate. I am not convinced that the `Sink` trait as is the correct design for this.

nikomatsakis (commented):
One rewording, otherwise, I think it's time to merge this.

nikomatsakis (approved):

cryptoquick:
Awesome! I like your last change, it helps. All LGTM!

cryptoquick:
> @cryptoquick rereading the code in this story: The connection is Clone, so there's no reason to re-wrap it with an Arc Mutex. You can just clone it as many times as you want and do whatever you'd do with a stream or a sink.
> 
> For other readers: I'm sorry this part spilled over here. The relevant bit is that "implement Sink" was the feature request, not "implement Clone"

I just got around to exploring this, and you're correct! I totally missed that trait. I think I'll contribute an update to this story with the changes reflected it this PR, and archive my fork with an explanation of how to use `tide-websockets` properly. It also has the advantage that it's much cleaner, and probably performs better, too! Fewer locks and atomic references.

https://github.com/ChainSafe/forest/pull/1089

I would have appreciated if this could have been pointed out earlier, however. It sure would have saved a lot of consternation. Did it take you a while to come to clarity on this? Even @yoshuawuyts and @skade never mentioned that solution. Admittedly, I did come off as a jerk, so I'll be forgiving.

jbr:
this is probably OT for the current repo, but I think we didn't get to a solution earlier on this issue because it was presented as a demand to implement a specific trait, instead of "how do I accomplish {goal}," and sending a link to another application is expecting an open source maintainer read your code in order to understand what you're _actually_ trying to do. It wasn't obvious why you thought you needed Sink, and starting a conversation by reopening asked and answered subjects didn't predispose anyone to put in the extra effort beyond "we're still not adding Sink."

cryptoquick:
Yeah, that's valid. I in my defense, I would say it's hard to know what you don't know.

nikomatsakis:
Well, I'm glad we found a good resolution in any case! <3

PR #100, opened 2021-03-26
Add status quo story about Grace trying to use gdb to step through a program

pnkfelix:
Written at tail end of group writing session hosted by niko, with useful input from other participants. (Especially point that Grace is the only character this really applies to...)

### DIFF
--- a/src/vision/grace_waits_for_gdb_next.md
+++ b/src/vision/grace_waits_for_gdb_next.md
@@ -58,7 +58,6 @@ Grace hits Ctrl-x a and gets a TUI mode view that includes this:
 │   57              let record = Box::new(Mutex::new(Record::new()));                                                                                                                                                                │
 │   58              let record = &*Box::leak(record);                                                                                                                                                                                │
 │   59                                                                                                                                                                                                                              
-|
 ```
 
 Excitedly Grace types `next` to continue to the next line of the function.
--- a/src/vision/grace_waits_for_gdb_next.md
+++ b/src/vision/grace_waits_for_gdb_next.md
@@ -1,0 +1,107 @@
+# 😱 Status quo stories: Template
+
+*This is a template for adding new "status quo" stories. To propose a new status quo PR, do the following:*
+
+* *Create a new file in the [`status_quo`] directory named something like `Alan_tries_to_foo.md` or `Grace_does_bar.md`, and start from [the raw source from this template]. You can replace all the italicized stuff. :)*
+* *Do not add a link to your story to the [`SUMMARY.md`] file; we'll do it after merging, otherwise there will be too many conflicts.*
+
+*For more detailed instructions, see the [How To Vision: Status Quo] page!*
+
+*If you're looking for ideas of what to write about, take a look at the [open issues]. You can also [open an issue of your own] to throw out an idea for others.*
+
+[How To Vision: Status Quo]: ../how_to_vision/status_quo.md
+[the raw source from this template]: https://raw.githubusercontent.com/rust-lang/wg-async-foundations/master/src/vision/status_quo/template.md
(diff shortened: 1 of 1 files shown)


src/vision/grace_waits_for_gdb_next.md
```diff
@@ -0,0 +1,119 @@
+# 😱 Status quo stories: Template
+
+*This is a template for adding new "status quo" stories. To propose a new status quo PR, do the following:*
```
nikomatsakis, line 3: Can you delete the italicized text?

src/vision/grace_waits_for_gdb_next.md
```diff
@@ -0,0 +1,119 @@
...
+Grace is now able to use `next` to walk through the main function. She does notice that the calls to `tokio::spawn` are skipped over by `next`, but that's not as much of a surprise to her, since those are indeed function calls that are taking async blocks. She sets breakpoints on the first line of each async block so that the debugger will stop when control reaches them as she steps through the code.
+
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+* **What are the morals of the story?**
+    * A common usage pattern: hitting `next` to go to what seems like the next statement, breaks down due to implementation details of `#[tokio::main]` and `async fn`.
+    * This is one example of where `next` break, in terms of what a user is likely to *want*. The other common scenario where the behavior of `next` is non-ideal is higher-order functions, like `option.and_then(|t| { ... }`, where someone stepping through the code probably *wants* `next` to set
```
nikomatsakis, line 101:
```suggestion
    * This is one example of where `next` breaks, in terms of what a user is likely to *want*. The other common scenario where the behavior of `next` is non-ideal is higher-order functions, like `option.and_then(|t| { ... }`, where someone stepping through the code probably *wants* `next` to set
```

nikomatsakis (changes requested):
Left a nit, but otherwise lgtm.

src/vision/grace_waits_for_gdb_next.md
```diff
@@ -0,0 +1,119 @@
...
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+Grace wants to walk through the behavior of a toy program.
+
+She first fires up `cargo run --verbose` to remind himself what the path to the target binary is. Part of the resulting Cargo output is:
```
rylev, line 30:
```suggestion
She first fires up `cargo run --verbose` to remind herself what the path to the target binary is. Part of the resulting Cargo output is:
```

rylev: Is this level of detail important? Can we just assume Grace knows where the debug binary is? I'm just not sure it adds much to the story.

rylev (approved):
Short and sweet. Looks good to me 😊

src/vision/grace_waits_for_gdb_next.md
pnkfelix: done

src/vision/grace_waits_for_gdb_next.md
pnkfelix: Well, when I was doing these steps to demonstrate the problem  during the writing session (and thus literally using `cargo run --verbose` myself, Niko said aloud "ugh why don't we have `cargo gdb`..."

To me, that's a sign that this level of detail *is* relevant.

Issue #78, opened 2021-03-21
Async-closures or reference-parameterized-closures-returning-futures


stuhood:
### Brief summary

Not having async closures can mean needing to resort to macros to create certain (fairly common) types of combinators (ones where a closure that returns a future runs into [lifetime issues](https://users.rust-lang.org/t/async-closure-lifetime-problem/42353)).

We experienced this in https://github.com/pantsbuild/pants/issues/11548, and I ended up writing a macro to put what would have been a parameter to a closure-returning-a-future or async-closure directly on the stack instead: https://github.com/pantsbuild/pants/pull/11759

### Optional details

* (Optional) Which [character(s)] would be the best fit and why?
    * [x] [Alan]: the experienced "GC'd language" developer, new to Rust
    * [ ] [Grace]: the systems programming expert, new to Rust
    * [ ] [Niklaus]: new programmer from an unconventional background
    * [x] [Barbara]: the experienced Rust developer
* (Optional) What are the key points or morals to emphasize?
    * Working reference-parameter-taking-closures-returning-futures would be great to have, but would require fairly advanced lifetimes (GATs?) to use even once they are possible (would likely be Barbara's choice). On the other hand, async-closures are less general, but likely to be easier to use (would likely be Alan's choice).

[character(s)]: https://rust-lang.github.io/wg-async-foundations/vision/characters.html
[project(s)]: https://rust-lang.github.io/wg-async-foundations/vision/projects.html
[Alan]: https://rust-lang.github.io/wg-async-foundations/vision/characters/alan.html
[Grace]: https://rust-lang.github.io/wg-async-foundations/vision/characters/grace.html
[Niklaus]: https://rust-lang.github.io/wg-async-foundations/vision/characters/niklaus.html
[Barbara]: https://rust-lang.github.io/wg-async-foundations/vision/characters/barbara.html

nikomatsakis:
@stuhood thanks for the example! I'd love to see a write-up dramatizing and spelling out some of the downsides of this macro approach you ended up with.

alsuren:
I can provide my own story. It's more of a saga, but contains two examples of this in two different projects. I would consider myself to be an Alan at the time, but it's a tale that involves a few other characters like Alan and Barbara. I will use real names and references, and once it's all written down we can trim it down and extract a persona story out of it.

[...]

PR #101, opened 2021-03-26
Borrowing arguments ("Alan writes a web framework")

alsuren:
This PR contains a write-up of my experiences from https://github.com/rust-lang/wg-async-foundations/issues/78#issuecomment-808193936 

There is a bit of solutionising towards the end. I can rip that out if you want:

> Alan longs to be able to say "this function takes an async function as a callback" (`fn register_handler(handler: impl async Fn(state: &mut State) -> Result<Response, Error>)`) and have Rust elide the lifetimes for him, like how they are elided for async functions.

It's also super-trimmed-down compared to my full experiences, and we could probably flesh out the morals a bit better.

### DIFF
--- a/src/vision/status_quo/alan_writes_a_web_framework.md
+++ b/src/vision/status_quo/alan_writes_a_web_framework.md
@@ -198,6 +198,8 @@ where
 
 Alan is still not sure whether it can be simplified.
 
+Later on, other developers on the project attempt to extend this approach to work with closures, but they encounter limitations in rustc that seem to make it not work ([rust-lang/rust#70263](https://github.com/rust-lang/rust/issues/70263)).
+
 When Alan sees another open source project struggling with the same issue, he notices that Barbara has helped them out as well. Alan wonders how many people in the community would be able to write `.to_async_borrowing()` without help.
 
 ## 🤔 Frequently Asked Questions
--- a/src/vision/status_quo/alan_writes_a_web_framework.md
+++ b/src/vision/status_quo/alan_writes_a_web_framework.md
@@ -1,4 +1,4 @@
-# 😱 Status quo stories: Template
+# 😱 Status quo stories: Alan writes a web framework
 
 
 [How To Vision: Status Quo]: ../how_to_vision/status_quo.md
--- a/src/vision/status_quo/alan_writes_a_web_framework.md
+++ b/src/vision/status_quo/alan_writes_a_web_framework.md
@@ -1,0 +1,50 @@
+# 😱 Status quo stories: Template
+
+*This is a template for adding new "status quo" stories. To propose a new status quo PR, do the following:*
+
+* *Create a new file in the [`status_quo`] directory named something like `Alan_tries_to_foo.md` or `Grace_does_bar.md`, and start from [the raw source from this template]. You can replace all the italicized stuff. :)*
+* *Do not add a link to your story to the [`SUMMARY.md`] file; we'll do it after merging, otherwise there will be too many conflicts.*
+
+*For more detailed instructions, see the [How To Vision: Status Quo] page!*
+
+*If you're looking for ideas of what to write about, take a look at the [open issues]. You can also [open an issue of your own] to throw out an idea for others.*
+
+[How To Vision: Status Quo]: ../how_to_vision/status_quo.md
+[the raw source from this template]: https://raw.githubusercontent.com/rust-lang/wg-async-foundations/master/src/vision/status_quo/template.md
+[`status_quo`]: https://github.com/rust-lang/wg-async-foundations/tree/master/src/vision/status_quo
+[`SUMMARY.md`]: https://github.com/rust-lang/wg-async-foundations/blob/master/src/SUMMARY.md
+[open issues]: https://github.com/rust-lang/wg-async-foundations/issues?q=is%3Aopen+is%3Aissue+label%3Astatus-quo-story-ideas
+[open an issue of your own]: https://github.com/rust-lang/wg-async-foundations/issues/new?assignees=&labels=good+first+issue%2C+help+wanted%2C+status-quo-story-ideas&template=-status-quo--story-issue.md&title=
+
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+*Write your story here! Feel free to add subsections, citations, links, code examples, whatever you think is best.*
+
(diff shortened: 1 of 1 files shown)


jbr:
A big plus one on this one. Not being able to describe the trait of async functions with non-static lifetimes in the arguments is by far the biggest pain in async rust right now. I'm excited to see how GATs improve this situation, but that's about solutions. Whole web framework designs might be different if mutable borrows could be passed into handler functions instead of owned types

src/vision/status_quo/alan_writes_a_web_framework.md
```diff
@@ -0,0 +1,62 @@
+# 😱 Status quo stories: Template
+
+*This is a template for adding new "status quo" stories. To propose a new status quo PR, do the following:*
```
nikomatsakis, line 3: Can you delete the italicized text?

src/vision/status_quo/alan_writes_a_web_framework.md
```diff
@@ -0,0 +1,62 @@
...
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+*Write your story here! Feel free to add subsections, citations, links, code examples, whatever you think is best.*
+
+## 🤔 Frequently Asked Questions
+
+[YouBuy](../projects/YouBuy.md) is written using an async web framework that predates the stabilization of async function syntax. When Alan joins the company, it is using async functions for its business logic, but can't use them for request handlers because the framework doesn't support it yet. It requires the handler's return value to be `Box<dyn Future<...>>`. Rather than switching YouBuy to a different web framework, Alan decides to contribute to the web framework himself.
```
nikomatsakis, line 32: Can you maybe give an example of what the code looks like when it returns `Box<dyn Future<...>.`? I guess people are writing things with `async move { }` blocks?

src/vision/status_quo/alan_writes_a_web_framework.md
```diff
@@ -0,0 +1,62 @@
...
+
+After a bit of a slog, he manages to make the web framework capable of using an `async fn` as an http request handler. He does this by making a wrapper function that boxes up the `impl Future`.
+
+It's still not fantastically ergonomic though. Because the web framework predates async function syntax, it requires you to take ownership of the request context (`fn handler(state: State)`) and return it alongside your response in the success/error cases (`Ok((state, response))` or `Err((state, error))`). This means that Alan can't use the `?` operator inside his http request handlers. Alan knows answer is to make another wrapper function so that the handler can take an `&mut` reference to `State` for the lifetime of the future, but Alan can't work out how to express this. He submits his pull-request upstream as-is, but it nags on his mind that he has been defeated.
```
nikomatsakis, line 36: It'd be great to show an example of what is required, and maybe include using the `?` operator (and the resulting error...)

src/vision/status_quo/alan_writes_a_web_framework.md
```diff
@@ -0,0 +1,62 @@
...
+
+Shortly afterwards, someone raises a bug about `?`, and a few other web framework contributors try to get it to work, but they also get stuck. When Alan tries it, the compiler diagnostics keep sending him around in circles. He can work out how to express the lifetimes for a function that returns a `Box<dyn Future + 'a>` but not an `impl Future` because of how where clauses are expressed. Alan longs to be able to say "this function takes an async function as a callback" (`fn register_handler(handler: impl async Fn(state: &mut State) -> Result<Response, Error>)`) and have Rust elide the lifetimes for him, like how they are elided for async functions.
```
nikomatsakis, line 38: If it is possible to give any examples, that'd be great

src/vision/status_quo/alan_writes_a_web_framework.md
```diff
@@ -0,0 +1,62 @@
...
+
+A month later, one of the contributors finds a forum comment by Barbara explaining how to express what the Alan is after (using higher-order lifetimes and a helper trait). They implement this and merge it.
```
nikomatsakis, line 40: Link?

nikomatsakis (commented):
Thanks! A few comments based on a quick read! The short version is that more examples would help bring the story to life for me.

src/vision/status_quo/alan_writes_a_web_framework.md
```diff
@@ -0,0 +1,62 @@
...
+
+When Alan sees another open source project struggling with the same issue, he notices that Barbara has helped them out as well.
+
+
+* **What are the morals of the story?**
+    * Callback-based APIs with async callbacks are a bit fiddly, because of the `impl Future` return type, but not insurmountable.
```
rylev, line 46: I'd love to have a bit more explicit description of "fiddly" here. Maybe this will be answered when you add examples above though...

rylev (changes requested):
There's definitely some really great insights here, but it looks like it's missing a bit of detail to really make the lessons obvious. Adding example errors and a bit more explanation of what exactly was confusing will go a long way.

src/vision/status_quo/alan_writes_a_web_framework.md
alsuren: `.map_err()` only works if you don't need to use state again (which you do, because you need to return it in the happy case: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a38f3e88718ef9c6ea14d2c5dd57ad6a). The best you can do is a macro that contains a match (`my_try!(state, result);` or something. It's like travelling back in time or something ;-).

Thanks for your feedback. I have put the PR back into draft, and I will try to add some fully-fleshed-out examples on Wednesday/Thursday.

tanriol:
You're missing the final problem here: the final `to_async_borrowing` works nicely for async functions, but not for closures because the closure type inference is not powerful enough ([rust-lang/rust70263](https://github.com/rust-lang/rust/issues/70263)), or at least the developers haven't managed to trick rustc into handling it yet.

src/vision/status_quo/alan_writes_a_web_framework.md
```diff
@@ -0,0 +1,222 @@
...
+            let body = serde_json::to_string(&prods).expect("Failed to serialize prods.");
+            let res = create_response(&state, StatusCode::OK, mime::APPLICATION_JSON, body);
+            Ok((state, res))
+        }
+        Err(e) => Err((state, e.into())),
+    }
+}
+```
+
+It's still not fantastically ergonomic though. Because the handler takes ownership of State and returns it in tuples in the result, Alan can't use the `?` operator inside his http request handlers. If he tries use `?` in a handler, like this:
```
nikomatsakis, line 91:
```suggestion
It's still not fantastically ergonomic though. Because the handler takes ownership of State and returns it in tuples in the result, Alan can't use the `?` operator inside his http request handlers. If he tries to use `?` in a handler, like this:
```

nikomatsakis (approved):
I feel like this is pretty good now! Let me review the outstanding comments, but I might just merge.

src/vision/status_quo/alan_writes_a_web_framework.md
nikomatsakis: I think it's fairly clear

PR #104, opened 2021-03-29
Add an async streams status quo for Barbara

jonathandturner:
This adds a status quo story for Barbara based on our experience using async streams in Nushell.

### DIFF
--- a/src/vision/status_quo/barbara_tries_async_streams.md
+++ b/src/vision/status_quo/barbara_tries_async_streams.md
@@ -1,0 +1,54 @@
+# 😱 Status quo stories: Barbara tries async streams
+
+## 🚧 Warning: Draft status 🚧
+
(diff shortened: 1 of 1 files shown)


rylev (commented):
This is definitely a very interesting story - Thank you! It certainly has a bit of overlap with #48. I wonder what if we can reword the morals in light of #48 existing so we have a clear take-away from this story other than what is already covered in #48.

nikomatsakis:
@jonathandturner that sounds helpful (using summary as morals)

jonathandturner:
Updated

nikomatsakis:
Thanks @jonathandturner!

rylev (approved):
Awesome! I think this really highlights the issues of *using* streams where as the other stream story we have is about *creating* streams. Really great!

PR #108, opened 2021-03-30
Status quo: add FAQ entries based on code review feedback to Barbara Carefully Dismisses Embedded Future.

jrvanwhy:
[Rendered](https://github.com/jrvanwhy/wg-async-foundations/blob/bcdef-faq/src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md)

This PR adds some of the conversations from the review of https://github.com/rust-lang/wg-async-foundations/pull/85 to the FAQ section of the Barbara Carefully Dismisses Embedded Future story.

I also fixed a typo earlier in the FAQ section that merged two bullets into one.


src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -246,6 +246,49 @@ instead. Here are some ways in which these APIs are lighter weight than a
       community and ecosystem.
     * [Niklaus] would really have struggled. If he asked for help, he probably
       would've gotten conflicting advice from the community.
+* **`Future` has a lot of features that Barbara's traits don't have -- aren't
+    those worth the cost?**
+    * `Future` has a lot of additional features that are nice-to-have. However,
```
nikomatsakis, line 251: I'd sort of like to know what these features are, can they be concisely listed?

nikomatsakis (changes requested):
Thanks, this is really helpful. Left one comment.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
jrvanwhy: I added a list. The list has a lot in common with Aaron Turon's [blog post](http://aturon.github.io/tech/2016/09/07/futures-design/) describing the initial design of `Future`.

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
```diff
@@ -246,6 +246,69 @@ instead. Here are some ways in which these APIs are lighter weight than a
...
+    * However, the code size impact of `Future` is a deal-breaker, and no number
+      of nice-to-have features can outweigh a deal-breaker. Barbara's traits
+      have every feature she *needs*.
+    * Using `Future` saves developer time relative to building your own async
+      abstractions. Developers can use the time they saved to minimize code size
+      elsewhere in the project. In some cases, this may result in a net decrease
+      in code size for the same total effort. However, code size reduction
+      efforts have diminishing returns, so projects that expect to optimize code
+      size regardless likely won't find the tradeoff beneficial.
+* **Is the code size impact of `Future` fundamental, or can the design be
```
rylev, line 281: The question isn't directly addressed in the answer (and maybe more design work needs to happen before it can be answered), but it doesn't feel like this is a *fundamental* tradeoff. If `Waker` could be tweaked to allow the code doing the waking to provide relevant context of who they are to the `Waker` perhaps this could be addressed. Thoughts?

jrvanwhy: The graph of possible execution flows for a future that is waiting on two events merges and diverges:

```
+-------+    +-------+
|Event A|    |Event B|
+-------+    +-------+
      |        |
      V        V
    +------------+
    |Waker::wake |
    +------------+
          |
          V
    +------------+
    |Future::poll|
    +------------+
       |      |
       V      V
+--------+  +--------+
|Handle A|  |Handle B|
+--------+  +--------+
```

As long as this merge exists, `Future::poll` will need to do work to determine what work triggered the poll. Callback-based systems transmit this implicitly, because the execution paths never merge.

I'm currently unsure whether it's possible to avoid the execution path merge without breaking backwards compatibility. It may be possible to add a `poll_with_context<T>()` method to `Future`, with a corresponding `wake_with_context<T>()` method on `Waker`, so that code paths for different events remain distinct.

nikomatsakis (approved):

src/vision/status_quo/barbara_carefully_dismisses_embedded_future.md
nikomatsakis: I feel like the existing answer suffices. I'm not sure exactly what would be worth saying to make it more detailed.

PR #115, opened 2021-04-02
Barbara builds an async executor

sticnarf:
A "status quo" story about my first experience about writing an async executor.

### DIFF
--- a/src/vision/status_quo/barbara_builds_an_async_executor.md
+++ b/src/vision/status_quo/barbara_builds_an_async_executor.md
@@ -9,7 +9,7 @@ This is a draft "status quo" story submitted as part of the brainstorming period
 
 Barbara wants to set priorities to the tasks spawned to the executor. However, she finds no existing async executor provides such a feature so she decided to build her own async executor.
 
-First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She decides to use it to build task schedulers. It is not the main part of the story. She plans for each working thread to have a loop which repeatedly gets a task from the deque and polls it.
+First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She decides to use it to build task schedulers. She plans for each working thread to have a loop which repeatedly gets a task from the deque and polls it.
 
 But wait, what should we put into those queues to represent each "task"?
 
--- a/src/vision/status_quo/barbara_builds_an_async_executor.md
+++ b/src/vision/status_quo/barbara_builds_an_async_executor.md
@@ -140,15 +140,15 @@ Barbara finished her initial implementation of the async executor. Despite there
 
 *Here are some standard FAQ to get you started. Feel free to add more!*
 
-* **What are the morals of the story?**
+### **What are the morals of the story?**
   * It is difficult to customize any of the current async executors (to my knowledge). To have any bit of special requirement forces building an async executor from scratch.
   * It is also not easy to build an async executor. It needs quite some exploration and is error-prone. [`async-task`](https://github.com/smol-rs/async-task) is a good attempt to simplify the process but it could not satisfy all kinds of needs of customizing the executor (it does not give you the chance to extend the task itself).
-* **What are the sources for this story?**
+### **What are the sources for this story?**
   * The story was from my own experience about writing a new thread pool supporting futures: https://github.com/tikv/yatp.
   * People may feel strange about why we want to set priorities for tasks. Currently, the futures in the thread pool are like user-space threads. They are mostly CPU intensive. But I think people doing async I/O may have the same problem.
-* **Why did you choose Barbara to tell this story?**
+### **Why did you choose Barbara to tell this story?**
   * At the time of the story, I had written Rust for years but I was new to the concepts for async/await like `Pin` and `Waker`.
-* **How would this story have played out differently for the other characters?**
+### **How would this story have played out differently for the other characters?**
   * People with less experience in Rust may be less likely to build their own executor. If they try, I think the story is probably similar.
 
 [character]: ../characters.md
--- a/src/vision/status_quo/barbara_builds_an_async_executor.md
+++ b/src/vision/status_quo/barbara_builds_an_async_executor.md
@@ -9,9 +9,9 @@ This is a draft "status quo" story submitted as part of the brainstorming period
 
 Barbara wants to set priorities to the tasks spawned to the executor. However, she finds no existing async executor provides such a feature so she decided to build her own async executor.
 
-First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She could use it to build task schedulers. It is not the main part of the story. Each working thread can get a task from the scheduler and poll it in loop.
+First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She decides to use it to build task schedulers. It is not the main part of the story. She plans for each working thread to have a loop which repeatedly gets a task from the deque and polls it.
 
-But wait, what is a "task"?
+But wait, what should we put into those queues to represent each "task"?
 
 At first, Barbara thought it must contain the `Future` itself and the additional priority which was used by the scheduler. So she first wrote:
 
@@ -32,7 +32,7 @@ pub fn poll_task(task: Task) {
 }
 ```
 
-"How to create a waker?" Barbara asked herself. Quickly, she found the `Wake` trait. Seeing the `wake` method takes an `Arc<Self>`, she realized the task in the scheduler should be stored in an `Arc`. After some thought, she thinks it makes sense because both the deque in the scheduler and the waker may hold a reference to the task.
+"How do I create a waker?" Barbara asked herself. Quickly, she found the `Wake` trait. Seeing the `wake` method takes an `Arc<Self>`, she realized the task in the scheduler should be stored in an `Arc`. After some thought, she realizes it makes sense because both the deque in the scheduler and the waker may hold a reference to the task.
 
 To implement `Wake`, the `Task` should contain the sender of the scheduler. She changed the code to something like this:
 
@@ -61,7 +61,7 @@ pub fn poll_task(task: Arc<Task>) {
 
 The code still needed some change because the `future` in the  `Arc<Task>` became immutable. 
 
-"Okay. I can guarantee `Task` is created from a `Pin<Box<Future>>` and I think the same future will not be polled concurrently in two threads. So let me bypass the safety checks." Barbara confidently used some `unsafe` blocks to make it compile.
+"Okay. I can guarantee `Task` is created from a `Pin<Box<Future>>`, and I think the same future won't be polled concurrently in two threads. So let me bypass the safety checks." Barbara changed the future to a raw pointer and confidently used some `unsafe` blocks to make it compile.
 
 ```rust
 pub struct Task {
@@ -80,7 +80,7 @@ pub fn poll_task(task: Arc<Task>) {
 }
 ```
 
-Luckily, a colleague of Barbara noticed something wrong. The `wake` method could be called multiple times so multiple copies of the task could exist in the scheduler. The scheduler might not work correctly because of it and a more severe problem was that multiple threads might get copies of the same task from the scheduler and cause a race in polling the future.
+Luckily, a colleague of Barbara noticed something wrong. The `wake` method could be called multiple times so multiple copies of the task could exist in the scheduler. The scheduler might not work correctly because of this. What's worse, a more severe problem was that multiple threads might get copies of the same task from the scheduler and cause a race in polling the future.
 
 Barbara soon got a idea to solve it. She added a state field to the `Task`. By carefully maintaining the state of the task, she could guarantee there are no duplicate tasks in the scheduler and no race can happen when polling the future.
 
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,161 @@
...
+                _ => break,
+            }
+        }
+    }
+}
+pub fn poll_task(task: Arc<Task>) {
+    let waker = Waker::from(task.clone());
+    let mut cx = Context::from_waker(&waker);
+    loop {
+        task.state.store(POLLING, Release);
```
tmandry, line 124: I got tripped up wondering why you didn't need to read before polling (it's because wake() prevents the task from appearing in the scheduler twice). You do state this in the prose above, but maybe a comment here would help as well.

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,161 @@
...
+
+```rust
+pub fn poll_task(task: Task) {
+    let waker = todo!();
+    let mut cx = Context::from_waker(&waker);
+    task.future.as_mut().poll(&mut cx);
+}
+```
+
+"How to create a waker?" Barbara asked herself. Quickly, she found the `Wake` trait. After reading the documentation, she realized the task in the scheduler should be stored in an `Arc`. And to implement `Wake`, the `Task` should also contain the sender of the scheduler. She changed the code to something like this:
```
tmandry, line 35: At this point I'm thinking about building an I/O reactor, so it might help to explain (here or above) that this is for compute tasks, not async I/O ones.

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,161 @@
...
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+## The story
+
+Barbara wants to set priorities to the tasks spawned to the executor. However, she finds no existing async executor provides such a feature so she decided to build her own async executor.
+
+First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She could use it to build task schedulers. It is not the main part of the story. Each working thread can get a task from the scheduler and poll it in loop.
```
tmandry, line 12: This paragraph says she _could_ use it to build a scheduler, and below the existence of the scheduler is assumed. Maybe this paragraph should say she _does_ use it to build a scheduler. Also, it could briefly name the operations of that scheduler so the reader knows what `send` is.

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,161 @@
...
+    ...
+    unsafe {
+        Pin::new_unchecked(&mut *task.future).poll(&mut cx);
+    }
+}
+```
+
+Luckily, a colleague of Barbara noticed something wrong. The `wake` method could be called multiple times so multiple copies of the task could exist in the scheduler. The scheduler might not work correctly because of it and a more severe problem was that multiple threads might get copies of the same task from the scheduler and cause a race in polling the future.
+
+Barbara soon got a idea to solve it. She added a state field to the `Task`. By carefully maintaining the state of the task, she could guarantee there are no duplicate tasks in the scheduler and no race can happen when polling the future:
```
tmandry, line 83: Just as a suggestion, maybe this paragraph can say something about how she spends a lot of effort in writing the code below (at least, it would take me a lot of effort). Then the reader is already focused on the pain of having to write this to satisfy such a simple requirement.

tmandry (commented):
Nicely written! I was able to follow along quite well and each step along the way made sense. I added a few comments that might help improve the story somewhat.

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,161 @@
...
+}
+```
+
+Barbara finished her initial implementation of the async executor. Despite there were a lot more possible optimizations, Barbara already felt it is a bit complex. She was also confused about why she needed to care so much about polling and waking while her initial requirement was just adding additional information to the task for customizing scheduling.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+* **What are the morals of the story?**
```
Stupremee, line 143: The style of the FAQ sections was changed in #118 / #119, using sub headings (###) instead of nested bullet lists.

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,161 @@
...
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+## The story
+
+Barbara wants to set priorities to the tasks spawned to the executor. However, she finds no existing async executor provides such a feature so she decided to build her own async executor.
+
+First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She could use it to build task schedulers. It is not the main part of the story. Each working thread can get a task from the scheduler and poll it in loop.
+
+But wait, what is a "task"?
```
nikomatsakis, line 14:
```suggestion
But wait, what should we put into those queues to represent each "task"?
```

I think this is what you mean?

src/vision/status_quo/barbara_builds_an_async_executor.md
nikomatsakis: It'd be helpful I think to say *why* the task should be in an arc. Presumably the idea is that the `Waker` will need to hold a reference to the task so that it can be scheduled; I guess it will need to also hold a reference to the queue or something like that?

nikomatsakis: Oh, I guess you said that last part ('the sender of the scheduler")

nikomatsakis: Oh, and I guess the `Arc` is partly because of the waker API...

src/vision/status_quo/barbara_builds_an_async_executor.md
nikomatsakis:
```suggestion
First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She decides to use it to build task schedulers. It is not the main part of the story. She plans for each working thread to have a loop which repeatedly gets a task from the deque and polls it.
```

nikomatsakis (approved):

nikomatsakis:
This story is really good. I left various comments, but I don't think it needs much.

nikomatsakis (changes requested):
(Marking as Changes Requested)

src/vision/status_quo/barbara_builds_an_async_executor.md
sticnarf: Ah yes, we can assume that she has made a task scheduler before building the executor.

src/vision/status_quo/barbara_builds_an_async_executor.md
sticnarf: Exactly. Thanks for clarifying it.

sticnarf:
> It'd be helpful I think to say why the task should be in an arc. Presumably the idea is that the Waker will need to hold a reference to the task so that it can be scheduled; I guess it will need to also hold a reference to the queue or something like that?

> Oh, and I guess the Arc is partly because of the waker API...

Yes, the API is the biggest hint. And it can take some while to think out why it needs an `Arc`. I changed a little of the story text to convey it better.

src/vision/status_quo/barbara_builds_an_async_executor.md
rylev: Nit: I'm not sure the sentence "It is not the main part of the story." is actually helpful. We could probably just remove it.

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,168 @@
...
+
+```rust
+pub fn poll_task(task: Task) {
+    let waker = todo!();
+    let mut cx = Context::from_waker(&waker);
+    task.future.as_mut().poll(&mut cx);
+}
+```
+
+"How to create a waker?" Barbara asked herself. Quickly, she found the `Wake` trait. Seeing the `wake` method takes an `Arc<Self>`, she realized the task in the scheduler should be stored in an `Arc`. After some thought, she thinks it makes sense because both the deque in the scheduler and the waker may hold a reference to the task.
```
rylev, line 35:
```suggestion
"How do I create a waker?" Barbara asked herself. Quickly, she found the `Wake` trait. Seeing the `wake` method takes an `Arc<Self>`, she realized the task in the scheduler should be stored in an `Arc`. After some thought, she realizes it makes sense because both the deque in the scheduler and the waker may hold a reference to the task.
```

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,168 @@
...
+    let waker = Waker::from(task.clone());
+    let mut cx = Context::from_waker(&waker);
+    task.future.as_mut().poll(&mut cx);
+//  ^^^^^^^^^^^ cannot borrow as mutable
+}
+```
+
+The code still needed some change because the `future` in the  `Arc<Task>` became immutable. 
+
+"Okay. I can guarantee `Task` is created from a `Pin<Box<Future>>` and I think the same future will not be polled concurrently in two threads. So let me bypass the safety checks." Barbara confidently used some `unsafe` blocks to make it compile.
```
rylev, line 64:
```suggestion
"Okay. I can guarantee `Task` is created from a `Pin<Box<Future>>`, and I think the same future won't be polled concurrently in two threads. So let me bypass the safety checks." Barbara changed the future to a raw pointer and confidently used some `unsafe` blocks to make it compile.
```

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,168 @@
...
+
+pub fn poll_task(task: Arc<Task>) {
+    ...
+    unsafe {
+        Pin::new_unchecked(&mut *task.future).poll(&mut cx);
+    }
+}
+```
+
+Luckily, a colleague of Barbara noticed something wrong. The `wake` method could be called multiple times so multiple copies of the task could exist in the scheduler. The scheduler might not work correctly because of it and a more severe problem was that multiple threads might get copies of the same task from the scheduler and cause a race in polling the future.
```
rylev, line 83:
```suggestion
Luckily, a colleague of Barbara noticed something wrong. The `wake` method could be called multiple times so multiple copies of the task could exist in the scheduler. The scheduler might not work correctly because of this. What's worse, a more severe problem was that multiple threads might get copies of the same task from the scheduler and cause a race in polling the future.
```

rylev (commented):
Awesome! I left a few small suggestions, but it is already looking great.

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,168 @@
...
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+## The story
+
+Barbara wants to set priorities to the tasks spawned to the executor. However, she finds no existing async executor provides such a feature so she decided to build her own async executor.
+
+First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She could use it to build task schedulers. It is not the main part of the story. Each working thread can get a task from the scheduler and poll it in loop.
```
nikomatsakis:
```suggestion
First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She could use it to build task schedulers. Each working thread can get a task from the scheduler and poll it in loop.
```

nikomatsakis (changes requested):
one more nit, per @rylev

src/vision/status_quo/barbara_builds_an_async_executor.md
```diff
@@ -0,0 +1,168 @@
...
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+## The story
+
+Barbara wants to set priorities to the tasks spawned to the executor. However, she finds no existing async executor provides such a feature so she decided to build her own async executor.
+
+First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She decides to use it to build task schedulers. It is not the main part of the story. She plans for each working thread to have a loop which repeatedly gets a task from the deque and polls it.
```
nikomatsakis, line 12:
```suggestion
First, Barbara found [crossbeam-deque](https://crates.io/crates/crossbeam-deque) provides work-stealing deques of good quality. She decides to use it to build task schedulers. She plans for each working thread to have a loop which repeatedly gets a task from the deque and polls it.
```

nikomatsakis (changes requested):

nikomatsakis (approved):

nikomatsakis:
I'm going to go ahead and merge this!

PR #119, opened 2021-04-02
Rework all FAQ Sections to use Headings instead of bullet lists

### LINKED ISSUE #118: convert FAQ entries from bulleted lists to sections
The FAQ entries today use nested lists, but those are annoying when you want to embed code blocks. A better choice would be use `###` sub-headings (or `####`, as the case may be). These would also be linkable. 

To make this happen you would need to basically riprep through the repo to convert the existing "Frequently Asked Questions" lists.

Stupremee:
Resolves #118


nikomatsakis:
You are wonderful.

nikomatsakis (approved):

PR #123, opened 2021-04-05
add mermaid

nikomatsakis:
This isn't complete, have to modify the CI to install mdbook-mermaid


Stupremee:
It would be nice to add a small comment somewhere mentioning that mermaid is enabled so that a new contributor knows that it's available for use. Probably in the How-To pages or templates.

nikomatsakis:
I don't know how to test this but I guess I can just merge it and see what happens

nikomatsakis:
I left the dummy graph in there so we can see if it's working :)

PR #122, opened 2021-04-05
add new stories to SUMMARY and fix links

nikomatsakis:

PR #124, opened 2021-04-05
Remove ignore directives from closing backtick blocks

eminence:
CC #122


nikomatsakis:
lol I suppose my search-and-replace was a bit hasty

nikomatsakis:
I wish I could find a way to make `mdbook test` stop running rust code blocks as tests

PR #133, opened 2021-04-08
Add story: "Niklaus wants to use GhostCell-like cell borrowing"

uazu:


ibraheemdev (commented):
Other than the short section under "The story", this reads more like a research paper than a status quo story to me. I doubt that Niklaus (a new programmer) would be familiar with all these details.

src/vision/status_quo/niklaus_wants_to_use_ghostcell.md
```diff
@@ -0,0 +1,149 @@
...
+Right now a `&mut Context<'_>` is passed, and so within `Context`
+would be the ideal place to hold a borrow on the cell owner.  However
+as far as Niklaus can see there are difficulties with all the current
+implementations:
+
+- GhostCell (or qcell::LCell) is the best available solution, because
+  it doesn't have any restrictions on how many runtimes might be
+  running or whatever.  But Rust insists that the lifetimes `<'id>` on
+  methods are explicit, so it seems like that would force a change to
+  the signature of `poll`, which would break the ecosystem.
```
nikomatsakis, line 52: It would be interesting to show what the signature *might* look like -- even though it would be invalid syntax.

src/vision/status_quo/niklaus_wants_to_use_ghostcell.md
```diff
@@ -0,0 +1,149 @@
...
+
+- The other `qcell` types (QCell, TCell and TLCell) have various
+  restrictions or overheads which might make them unsuitable as a
+  general-purpose solution in the standard library.  However they do
+  have the positive feature of not requiring any change in the
+  signature of `poll`.  It looks like they could be added to `Context`
+  without breaking anything.
```
nikomatsakis, line 59: I don't know what these types are -- it would be nice to give some examples, maybe? e.g., show how he is able to use them?

src/vision/status_quo/niklaus_wants_to_use_ghostcell.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Proof of concept
+
+The [Stakker](https://crates.io/crates/stakker) runtime makes use of
+qcell-based statically-checked cell borrowing.  It uses this to get
+zero-cost access to actors, guaranteeing at compile time that no
+actor can access any other actor's state.  It also uses it to allow
+inter-actor [shared
+state](https://docs.rs/stakker/0.2.1/stakker/struct.Share.html) to be
+accessed safely and zero-cost, without RefCell.
```
nikomatsakis, line 73: examples!

src/vision/status_quo/niklaus_wants_to_use_ghostcell.md
```diff
@@ -0,0 +1,149 @@
...
+
+Switching implementations like this doesn't seem like an option for
+the standard library.
+
+### Way forward
+
+If the feature is of interest, it needs some brain-time from someone
+to see if there's any way to add this.  For example could the compiler
+derive the `<'id>` annotations automatically for GhostCell?  Or is
+there any other way of making this work in the standard library?
```
nikomatsakis, line 89: This isn't very narrative, but maybe that's ok. I think it might be fine to dress it up as Niklaus trying out various approaches and getting an idea of their pros/cons.

nikomatsakis (changes requested):
I enjoyed this! I agree it's quite different from the other stories. I think I would write it in terms of Barbara -- I feel like a Rust enthusiast would quite likely be very into qcell -- this isn't quite how i imagined Niklaus. Specifically I had imagined a character that is more 'new to programming'. However, I'm ultimately happy to let you decide your own character. :) 

I don't really have any specific suggestions. I am curious to read more about qcell.

Ordinarily I like to suggest a very narrative form but I'm not sure how helpful it would be for this particular story. I think the story does a good job of motivating the appeal of ghost cells to Niklaus (or Barbara, or whomever) and since he is mostly just contemplating how he could use them, there isn't *that* much to tell. I left some ideas where I thought things could be made clearer, mostly by adding some examples.

uazu:
Okay, I'll change to Barbara. I will also expand it out where requested -- and try and adjust to fit the "story" style!

uazu:
I've tried to address the comments. I'm happy to edit further as necessary.

src/vision/status_quo/barbara_wants_to_use_ghostcell.md
```diff
@@ -0,0 +1,274 @@
...
+
+If you would like to expand on this story, or adjust the answers to
+the FAQ, feel free to open a PR making edits (but keep in mind that,
+as they reflect peoples' experiences, status quo stories [cannot be
+wrong], only inaccurate). Alternatively, you may wish to [add your own
+status quo story][htvsq]!
+
+## The story
+
+Barbara quite likes using statically-checked cell borrowing.  This is
```
rylev, line 18: It took me a second to figure out that the cell is the first sentence refers to the cells like `std::cell::{Cell, RefCell}`. 

Perhaps you could rephrase this to "Barbara would like to use a form of mutable memory location like `std::cell::Cell` that are statically checked and guaranteed to not panic at runtime".

src/vision/status_quo/barbara_wants_to_use_ghostcell.md
```diff
@@ -0,0 +1,274 @@
...
+implemented in various ways in
+[GhostCell](http://plv.mpi-sws.org/rustbelt/ghostcell/) and
+[`qcell`](https://docs.rs/qcell/0.4.1/qcell/).  She would like to use
+statically-checked cell borrowing within futures, but there is no way
```
rylev, line 22: It would be good to show why there is no way to get the "owner borrow" through the `Future::poll` call here. This is explained a little bit further down, but I was confused for a little while until I got to that part.

rylev (commented):
I'm not sure how I feel about this being a status quo story - The first half motivates the use of the statically checked cells (though the motivation could be expanded IMO), but then the second half suggests fixes to the standard library which seems more appropriate for a shiny future.

As the story is written it's more of an intellectual exercise than a story motivated by a real world use case. That being said, I am absolutely sure there are many good real-world use cases for this, I just think the story should better reflect that.

In other words, status quo stories are meant to highlight the issues with async Rust and not really to propose solutions (which will be the job of shiny futures). I'd like to see this story focus more on the problem space than the solution space. 

I think this story would be better if were rephrased to more extensively motivate statically checked cells and have slightly less detail on a proposal that could go into the standard library. Then some of that detail can be moved to a "shiny future" story.

src/vision/status_quo/barbara_wants_to_use_ghostcell.md
uazu: I think "cell" is the correct terminology in the Rust world. If I Google "rust cell" I get Cell, RefCell, UnsafeCell etc. However I will add a clarification in a separate sentence.

src/vision/status_quo/barbara_wants_to_use_ghostcell.md
uazu: Okay, I will expand that part

uazu:
> I think this story would be better if were rephrased to more extensively motivate statically checked cells and have slightly less detail on a proposal that could go into the standard library. Then some of that detail can be moved to a "shiny future" story.

I'm not sure I'm the right person to fully motivate this, and probably nor is Barbara -- although I could try. If you want a "RefCell tale of woe" then perhaps we need a more naive character. I assume that Barbara knows very well the problems of RefCell borrows and avoids those problems through diligent manual checking. I avoid `RefCell::borrow_mut` as much as I can, so I don't have a tale of woe either. Just recently, coding with it I realized that I had to wake up and pay attention and double-check my code, and actually I found one place where I needed to drop the borrow in order to avoid the possibility of a panic. But this is prevention. Someone without that awareness could easily get a lot deeper in before having unexpected panics and then having to debug and so on. But I don't have a story like that because I'm already aware of the danger.

Anyway I will expand the story part and explain more stuff and see if I can motivate it better. Regarding splitting it up -- really this is for you guys to decide. Niko asked me to expand on the qcell stuff, so I did. I don't know how best to fit into your structure. I was running with "Barbara has a curious mind and tries to solve the problem herself". Whether there is a bright future, I'm not sure. It may be technically impossible given the constraints, but Barbara (and I) can't be sure. Maybe it IS possible, and if someone can see that bright future path then that would be fantastic!

nikomatsakis:
> I think this story would be better if were rephrased to more extensively motivate statically checked cells and have slightly less detail on a proposal that could go into the standard library. Then some of that detail can be moved to a "shiny future" story.

Hmm, @rylev, I think i agree. What i could imagine would be a story that highlights a case where the user has to debug panics, which ghostcell would avoid. and perhaps clarifies why cells are needed in the first place (this might reference other stories).

uazu:
Thinking some more about the motivation, I don't know whether this is a practical problem for lots of people or not. Is there anything special about how async/await code is structured that means you can be lax with RefCell borrows and it all works out fine anyway? In my runtime glue code, there was definitely a case where I could have had a panic. But if the problem is limited to coders of runtimes it is not such a big deal.

nikomatsakis:
I think that there will be more cells in async code than other code specifically because of the fact that `&mut` contexts don't place nicely, as described in [alan lost the world](https://github.com/rust-lang/wg-async-foundations/blob/master/src/vision/status_quo/alan_lost_the_world.md0. I have no idea if ghostcell can help here, though!

uazu:
I can accept that the motivation is not strong enough right now, unless it happens to connect with another story.

nikomatsakis:
@uazu My personal policy is to try and merge all stories -- this is brainstorming, after all, not making any full decisions! So, I would like to merge this PR. However, before doing so, it'd be good if we make as strong a case as we can. Editing the story to focus a bit on the practical difficulties, if we can think of some, would be excellent. As I said, I think that we are likely to have cells showing up to share context between tasks.

uazu:
I've added two examples. If anyone can think of other example scenarios, I can add them.

nikomatsakis (approved):
Seems good to me, thanks!

uazu:
As this hasn't been merged yet, I thought it best to correct a small error in example 1. I have now constructed the full example and it does panic. I didn't just paste in the original code as it is a lot more complicated and would take more explaining. The example now illustrates the problem situation better.

nikomatsakis:
Great! Merging.

Issue #82, opened 2021-03-22
Using an async block or function in a trait


seanmonstar:
### Brief summary

It's very easy to write inherent `async fn`s on a type, but if using them inside a trait method is much harder. For example, this pseudo code doesn't work, and making it work isn't intuitive:

```rust
impl Service for Handler {
    type Future = WhatTypeGoesHere;
    fn call(&self, req: Request) -> Self::Future {
        async move {
            self.count(req).await;
            Ok(self.construct_response())
        }
    }
}
```

### Optional details

* (Optional) Which [character(s)] would be the best fit and why?
    * Any of the characters, since it's hard in general. [Barbara][] may know how to work around it, but she likely still doesn't like it.
* (Optional) What are the key points or morals to emphasize?
    * It's confusing that you can't write `async fn` when implementing a trait.
    * It's hard to return a future from a trait method that came from another `async fn` or `async` block.

[character(s)]: https://rust-lang.github.io/wg-async-foundations/vision/characters.html
[project(s)]: https://rust-lang.github.io/wg-async-foundations/vision/projects.html
[Alan]: https://rust-lang.github.io/wg-async-foundations/vision/characters/alan.html
[Grace]: https://rust-lang.github.io/wg-async-foundations/vision/characters/grace.html
[Niklaus]: https://rust-lang.github.io/wg-async-foundations/vision/characters/niklaus.html
[Barbara]: https://rust-lang.github.io/wg-async-foundations/vision/characters/barbara.html

PR #134, opened 2021-04-08
status-quo story: Alan needs async in traits

zeenix:
Result of the [async traits session](https://hackmd.io/TTD9uVNxQYKOF_5omKxogw) held on Apr 8, 2021.

Fixes https://github.com/rust-lang/wg-async-foundations/issues/82

### DIFF
--- a/src/vision/status_quo/alan_needs_async_in_traits.md
+++ b/src/vision/status_quo/alan_needs_async_in_traits.md
@@ -1,0 +1,110 @@
+# 😱 Status quo stories: Alan needs async in traits
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+Alan is working on a project with Barbara which has already gotten off to a [somewhat rocky start](./barbara_anguishes_over_http.md). He is working on abstracting away the HTTP implementation the library uses so that users can provide their own. He wants the user to implement an async trait called `HttpClient` which has one method `perform(request: Request) -> Response`. Alan tries to create the async trait:
+
+```rust
+trait HttpClient {
+    async fn perform(request: Request) -> Response;
+}
+```
+
+When Alan tries to compile this, he gets an error:
+
+```
+ --> src/lib.rs:2:5
+  |
+2 |     async fn perform(request: Request) -> Response;
+  |     -----^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  |     |
+  |     `async` because of this
+  |
+  = note: `async` trait functions are not currently supported
+  = note: consider using the `async-trait` crate: https://crates.io/crates/async-trait
+```
+
+Alan, who has been using Rust for a little while now, has learned to follow compiler error messages and adds `async-trait` to his `Cargo.toml`. Alan follows the README of `async-trait` and comes up with the following code:
+
+```rust
+#[async_trait]
+trait HttpClient {
+    async fn perform(request: Request) -> Response;
+}
+```
+
+Alan's code now compiles, but he also finds that his compile times have gone from under a second to around 6s, at least for a clean build.
+
+After Alan finishes adding the new trait, he shows his work off to Barbara and mentions he's happy with the work but is a little sad that compile times have worsened. Barbara, an experienced Rust developer, knows that using `async-trait` comes with some additional issues. In this particular case she is especially worried about tying their public API to a third-party dependency. Even though it is technically possible to implement traits annotated with `async_trait` without using `async_trait`, doing so in practice is very painful. For example `async_trait`:
+
+* handles lifetimes for you if the returned future is tied to the lifetime of some inputs.
+* boxes and pins the futures for you.
+
+which the implementer will have to manually handle if they don't use `async_trait`. She decides to not worry Alan with this right now. Alan and Barbara are pretty happy with the results and go on to publish their crate which gets lots of users.
+
+Later on, a potential user of the library wants to use their library in a `no_std` context where they will be providing a custom HTTP stack. Alan and Barbara have done a pretty good job of limiting the use of standard library features and think it might be possible to support this use case. However, they quickly run into a show stopper: `async-trait` boxes all of the futures returned from a async trait function. They report this to Alan through an issue.
+
+Alan, feeling (over-) confident in his Rust skills, decides to try to see if he can implement async traits without using `async-trait`. 
+
+```rust 
+trait HttpClient {
+   type Response: Future<Output = Response>;
+
+   fn perform(request: Request) -> Self::Response; 
+}
+```
+
+Alan seems to have something working, but when he goes to update the examples of how to implement this trait in his crate's documentation, he realizes that he either needs to:
+
+* use trait object:
+
+  ```rust
+  struct ClientImpl;
+
+  impl HttpClient for ClientImpl {
+      type Response = Pin<Box<dyn Future<Output = Response>>>;
+
+      fn perform(request: Request) -> Self::Response {
+          Box::pin(async move {
+              // Some async work here creating Reponse
+          })
+	  }
+  }
+  ```
+
+  which wouldn't work for `no_std`.
+
+* implement `Future` trait manually, which isn't particulary easy/straight-forward for non-trivial cases, especially if it involves making other async calls (likely).
+
+After a lot of thinking and discussion, Alan and Barbara accept that they won't be able to support `no_std` users of their library and add mention of this in crate documentation.
+
+## 🤔 Frequently Asked Questions
+
+### **What are the morals of the story?**
+
+* `async-trait` is awesome, but has some drawbacks
+    * compile time increases
+    * performance cost of boxing and dynamic dispatch
+    * not a standard solution so when this comes to language, it might break things
+* Trying to have a more efficient implementation than `async-trait` is likely not possible. 
+
+### **What are the sources for this story?**
+
+* [Zeeshan](https://github.com/zeenix/) is looking for a way to implement async version of the [service-side zbus API](https://docs.rs/zbus/1.9.1/zbus/trait.Interface.html).
+* [Ryan](https://github.com/rylev) had to use `async-trait` in an internal project.
+
+### **Why did you choose Alan to tell this story?**
+
+We could have used Barbara here but she'd probably know some of the work-arounds (likely even the details on why they're needed) and wouldn't need help so it wouldn't make for a good story. Having said that, Barbara is involved in the story still so it's not a pure Alan story.
+
+### **How would this story have played out differently for the other characters?**
+
+* Barbara: See above.
+* Grace: Probably won't know the solution to these issues much like Alan, but might have an easier time understanding the **why** of the whole situation. 
+* Niklaus: would be lost - traits are somewhat new themselves. This is just more complexity, and Niklaus might not even know where to go for help (outside of compiler errors).
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -32,6 +32,7 @@
       - [Alan started trusting the Rust compiler, but then... async](./vision/status_quo/alan_started_trusting_the_rust_compiler_but_then_async.md)
       - [Alan tries to debug a hang](./vision/status_quo/alan_tries_to_debug_a_hang.md)
       - [Alan writes a web framework](./vision/status_quo/alan_writes_a_web_framework.md)
+      - [Alan needs async in traits](./vision/status_quo/alan_needs_async_in_traits.md)
       - [Barbara anguishes over HTTP](./vision/status_quo/barbara_anguishes_over_http.md)
       - [Barbara carefully dismisses embedded `Future`](./vision/status_quo/barbara_carefully_dismisses_embedded_future.md)
       - [Barbara makes their first foray into async](./vision/status_quo/barbara_makes_their_first_steps_into_async.md)
--- a/src/vision/status_quo/alan_tries_to_debug_a_hang.md
+++ b/src/vision/status_quo/alan_tries_to_debug_a_hang.md
@@ -130,7 +130,7 @@ Disgruntled, Alan begins the arduous, boring task of instrumenting the applicati
 * Some of the characters with either a background in Rust or a background in systems programming might know that Rust's async doesn't always map to an underlying system feature and so they might expect that `gdb` or `lldb` is unable to help them.
 * Barbara, the experienced Rust dev, might also have used a tracing/instrumentation library from the beginning and have that to fall back on rather than having to do the work to add it now.
 
-[character]: ../characters.md
+[characters]: ../characters.md
 [status quo stories]: ./status_quo.md
 [Alan]: ../characters/alan.md
 [Grace]: ../characters/grace.md


src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+  |     |
+  |     `async` because of this
+  |
+  = note: `async` trait functions are not currently supported
+  = note: consider using the `async-trait` crate: https://crates.io/crates/async-trait
+```
+
+Alan, using Rust for a little while now, has learned to follow compiler error messages and adds `async-trait` to the dependencies list. Alan follows the README of `async-trait` and gets the code to compile. Alan is somewhat dismayed that the compile times have increased considerably after adding this new dependency.
+
+**TODO::** Maybe add representive pre and post trait compile times here.
```
Stupremee, line 35: I think a small and simple mention like "the compile times doubled" or "went from 10s to 20s" is good

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+    * lack of control over whether futures are `Send` or `Sync`
+    * not a standard solution so when this comes to language, it might break things
+* Trying to have a more efficient implementation than `async-trait` is likely not possible. 
+
+### **What are the sources for this story?**
+
+* [Zeeshan](https://github.com/zeenix/) is looking for a way to implement async version of the [service-side zbus API](https://docs.rs/zbus/1.9.1/zbus/trait.Interface.html).
+* [Ryan](https://github.com/rylev) had to use `async-trait` in an internal project.
+
+## **Why did you choose Alan to tell this story?**
```
Stupremee, line 93: There are only two #

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+          })
+	  }
+  }
+  ```
+
+  which wouldn't work for `no_std`.
+
+* implement `Future` trait manually, which isn't particulary easy/straight-forward for non-trivial cases, especially if it involves making other async calls (likely).
+
+**TODO::** Would make sense for Alan to reach out to Barbara at this point and she should be able to give some good advice.
```
nikomatsakis, line 75: There isn't *that* much he can do, right? Maybe I don't know the tricks =)

nikomatsakis (changes requested):

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+
+## 🤔 Frequently Asked Questions
+
+### **What are the morals of the story?**
+
+* `async-trait` is awesome, but has some drawbacks
+    * compile time increases
+    * performance cost of boxing and dynamic dispatch
+    * lack of control over whether futures are `Send` or `Sync`
```
ibraheemdev, line 84: `async-trait` *does* give you control over this. You can use `#[async_trait(?Send)]` if you don't want to add `Send` bounds.

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+2 |     async fn perform(request: Request) -> Response;
+  |     -----^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  |     |
+  |     `async` because of this
+  |
+  = note: `async` trait functions are not currently supported
+  = note: consider using the `async-trait` crate: https://crates.io/crates/async-trait
+```
+
+Alan, using Rust for a little while now, has learned to follow compiler error messages and adds `async-trait` to the dependencies list. Alan follows the README of `async-trait` and gets the code to compile. Alan is somewhat dismayed that the compile times have increased considerably after adding this new dependency.
```
ibraheemdev, line 33:
```suggestion
Alan, who has been using Rust for a little while now, has learned to follow compiler error messages and adds `async-trait` to his `Cargo.toml`. Alan follows the README of `async-trait`, and comes up with the following code:
'''rust
#[async_trait]
trait HttpClient {
    async fn perform(request: Request) -> Response;
}
'''
Despite getting his code to compile, Alan is somewhat dismayed that the compile times have increased considerably after adding this new dependency.
```
(markdown doesn't like nested code blocks so I used apostrophes instead of backticks - don't click "apply suggestion" :)

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+* implement `Future` trait manually, which isn't particulary easy/straight-forward for non-trivial cases, especially if it involves making other async calls (likely).
+
+**TODO::** Would make sense for Alan to reach out to Barbara at this point and she should be able to give some good advice.
+
+## 🤔 Frequently Asked Questions
+
+### **What are the morals of the story?**
+
+* `async-trait` is awesome, but has some drawbacks
+    * compile time increases
```
ibraheemdev, line 82: Is the compile time increase really that bad?

ibraheemdev (changes requested):

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: I too wondered about this...

src/vision/status_quo/alan_needs_async_in_traits.md
Stupremee: She might suggest writing a custom proc-macro that is like async-trait, but has a special no_std mode which allows using it without std? Alan as a newbie in Rust doesn't know much about proc-macros and how to write them himselfes. It's also a huge project for a Rust beginner, but for Barbara as an advanced Rust Developer, she probably has experience with proc-macros and can help out to write it.

Another tip she might give, is how to implement the `Future` trait properly and how to use / create helpers like the `std::task::ready` macro and other stuff.

ibraheemdev: There is the `poll_x` pattern, and I saw a [cool crate recently that can help with that](https://crates.io/crates/async-trait-ext).... but I guess that falls under implementing the `Future` manually/with a different proc-macro.

nikomatsakis: The only way to make things work with no-std that I know of is to rely on nightly only features, is the thing.

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+Alan is working on a project with Barbara which has already gotten off to a [somewhat rocky start](https://github.com/rust-lang/wg-async-foundations/pull/111). He is working on abstracting away the HTTP implementation the library uses so that users can provide their own. He wants the user to implement an async trait called `HttpClient` which has one method `perform(request: Request) -> Response`. Alan tries to create the async trait:
```
eminence, line 11: Let's link the to merged story, not the PR:

```suggestion
Alan is working on a project with Barbara which has already gotten off to a [somewhat rocky start](./barbara_anguishes_over_http.md). He is working on abstracting away the HTTP implementation the library uses so that users can provide their own. He wants the user to implement an async trait called `HttpClient` which has one method `perform(request: Request) -> Response`. Alan tries to create the async trait:
```

src/vision/status_quo/alan_needs_async_in_traits.md
eminence: I did a quick test with two dummy projects.

For the code using async-trait, it looks like this:

```rust
#[async_trait]
trait TraitA {
    async fn A();
}

#[async_trait]
impl TraitA for () {
    async fn A() {
        todo!()
    }
}
```

For the code without async-trait, it looks like:

```rust
trait TraitA {
    type R: Future<Output = ()>;
    fn A() -> Self::R;
}


impl TraitA for () {
    type R = Pin<Box<dyn Future<Output = ()>>>;
    fn A() -> Self::R {
        Box::pin(async move {
            todo!()
        })
    }
}
```

I duplicated these traits a total of 7 times (so I have `traitA` through `traitF`)

Using a build script that simply touches main.rs, hyperfine says:

```
$ hyperfine ./foo-async-trait/build.sh ./foo/build.sh
Benchmark #1: ./foo-async-trait/build.sh
  Time (mean ± σ):     609.0 ms ±  11.9 ms    [User: 500.4 ms, System: 125.1 ms]
  Range (min … max):   591.9 ms … 623.3 ms    10 runs

Benchmark #2: ./foo/build.sh
  Time (mean ± σ):     505.7 ms ±   5.1 ms    [User: 398.4 ms, System: 122.5 ms]
  Range (min … max):   499.2 ms … 514.5 ms    10 runs

Summary
  './foo/build.sh' ran
    1.20 ± 0.03 times faster than './foo-async-trait/build.sh'
```

ibraheemdev: That doesn't seem like *that* big of a deal. It's also worth noting that many crates will already have `async-trait`'s deps (`proc-macro`, `syn`, `quote`) as a transitive dep from elsewhere.

src/vision/status_quo/alan_needs_async_in_traits.md
rylev: FWIW adding `async-trait` to an empty project brings clean build times from ~0.4s to ~5.75s on my machine.

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+  |
+  = note: `async` trait functions are not currently supported
+  = note: consider using the `async-trait` crate: https://crates.io/crates/async-trait
+```
+
+Alan, using Rust for a little while now, has learned to follow compiler error messages and adds `async-trait` to the dependencies list. Alan follows the README of `async-trait` and gets the code to compile. Alan is somewhat dismayed that the compile times have increased considerably after adding this new dependency.
+
+**TODO::** Maybe add representive pre and post trait compile times here.
+
+After Alan finishes adding the new trait, he shows his work off to Barbara and mentions he's happy with the work but is a little sad that compile times have worsened. Barbara, an experienced Rust developer, knows that using `async-trait` comes with some issues and in this case is especially worried about tying their public API to a third-party dependency. She decides to not worry Alan with this right now. Alan and Barbara are pretty happy with the results and go on to publish their crate which gets lots of users.
```
rylev, line 37:
```suggestion
After Alan finishes adding the new trait, he shows his work off to Barbara and mentions he's happy with the work but is a little sad that compile times have worsened. Barbara, an experienced Rust developer, knows that using `async-trait` comes with some additional issues. In this particular case she is especially worried about tying their public API to a third-party dependency. Even though it is technically possible to implement traits annotated with `async_trait`, doing so in practice is very painful. She decides to not worry Alan with this right now. Alan and Barbara are pretty happy with the results and go on to publish their crate which gets lots of users.
```

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+
+   fn perform(request: Request) -> Self::Response; 
+}
+```
+
+Alan seems to have something working, but when he goes to update the examples of how to implement this trait in his crate's documentation, he realizes that he either needs to:
+
+* use trait object:
+
+  ```rust,ignore
```
rylev, line 57: nit: I don't think ignore is necessary here since none of the code samples are run.

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+
+Later on, a potential user of the library wants to use their library in a `no_std` context where they will be providing a custom HTTP stack. Alan and Barbara have done a pretty good job of limiting the use of standard library features and think it might be possible to support this use case. However, they quickly run into a show stopper: `async-trait` boxes all of the futures returned from a async trait function. They report this to Alan through an issue.
+
+Alan, feeling (over-) confident in his Rust skills, decides to try to see if he can implement async traits without using `async-trait`. 
+
+```rust 
+struct Request(String);
+
+trait HttpClient {
+   type Response: Future<Output = ()>;
```
rylev, line 47:
```suggestion
   type Response: Future<Output = Response>;
```

src/vision/status_quo/alan_needs_async_in_traits.md
rylev: I would simply have Barbara mention exactly what Niko said - it's likely not possible without using nightly-only features. We can then have compiling on stable as a part of their requirements, leading this to be a dead end.

src/vision/status_quo/alan_needs_async_in_traits.md
rylev: Did you test clean build times? Considering that this would likely not impact build times _at all_ if it were a built in language feature, I think having build times impacted at all is a concern that should be taken into account.

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -0,0 +1,101 @@
...
+	  }
+  }
+  ```
+
+  which wouldn't work for `no_std`.
+
+* implement `Future` trait manually, which isn't particulary easy/straight-forward for non-trivial cases, especially if it involves making other async calls (likely).
+
+**TODO::** Would make sense for Alan to reach out to Barbara at this point and she should be able to give some good advice.
+
```
rylev, line 76: Let's add a final sentence to wrap this up. Something like "Alan and Barbara accept that they won't be able to support `no_std` users of their library."

rylev (commented):
Looking good! It's almost there

rylev:
I tied this PR to https://github.com/rust-lang/wg-async-foundations/issues/82

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: Ah, I saw `ignore` being used in other parts, so I added to be sure. :)

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: @rylev are you sure about this? `Response` associated type is a `Future` that gives `Response` type?

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: and that's too big an impact, for it to be a non-concern.

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: > Even though it is technically possible to implement traits annotated with `async_trait`, doing so in practice is very painful

We should describe at least one example of a pain point here IMO (IOW need to be a bit more specific).

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: What I was hoping for here is that some of you might know of some API that **somehow** allows you to turn an `impl Future` into a named future, similar to how [`futures::stream::unfold`](https://docs.rs/futures/0.3.13/futures/stream/fn.unfold.html) allows you to create a stream out of an `impl Future` but I forgot that will [require boxing](https://gitlab.freedesktop.org/dbus/zbus/-/blob/491e5456d66510110ee8e730f3bb9e5a692844c0/zbus/src/azync/proxy.rs#L310) as well.

As for a proc-macro way, from what I understood from the [README](https://crates.io/crates/async-trait-ext), `async-trait-ext` only requires a nightly feature for provided methods in traits. Alan doesn't necessarily need provided methods so he can possibly make use of `async-trait-ext` w/o nightly, can't he?

src/vision/status_quo/alan_needs_async_in_traits.md
rylev: I think some pain points are: 
* `async_trait` handles lifetimes for you if the returned future is tied to the lifetime of some inputs
* `async_trait` also boxes and pins the futures for you.

src/vision/status_quo/alan_needs_async_in_traits.md
rylev: Yes, we can change the name of the associated type, but the future will eventually yield an HTTP response

ibraheemdev: Yeah, that is the equivalent to `async fn perform(request: Request) -> Response;`

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: It used to be require, but we stopped running `mdbook test`

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: +1 for having a code example.

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: Can we include these numbers? That's very interesting.

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: Maybe include @rylev's numbers:

---

```suggestion
Alan, who has been using Rust for a little while now, has learned to follow compiler error messages and adds `async-trait` to his `Cargo.toml`. Alan follows the README of `async-trait` and comes up with the following code:
'''rust
#[async_trait]
trait HttpClient {
    async fn perform(request: Request) -> Response;
}
'''
Alan's code now compiles, but he also finds that his compile times have gone from under a second to around 6s, at least for a clean build.
```

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: I took a stab at that above :)

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: I agree it would be nice to have some examples, or a link to them -- I was very curious what this meant.

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: I'd say we should remove this bullet

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis:
```suggestion
### **Why did you choose Alan to tell this story?**
```

nikomatsakis (changes requested):
Very close! Some suggestions.

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: thinking more about it, it's actually not `no_std` that's impossible here but rather `no_alloc`, although I'm not sure if this typically makes any difference in the real world.

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: @ibraheemdev thanks, yeah I know but just got confused because @rylev chose the same name for the output type as the associated type but he explained so all good. :)

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: Thanks @nikomatsakis.

src/vision/status_quo/alan_needs_async_in_traits.md
zeenix: @nikomatsakis I added the pain points @rylev listed above. Do we need more?

zeenix:
Thanks for the reviews folks. PTALA. Only one unresolved thread left (although I think it's likely addressed). Once all resolved, I'd like to rebase to resolve the conflict and to squash all the `fixup` commits to the right commit.

src/vision/status_quo/alan_needs_async_in_traits.md
```diff
@@ -30,21 +30,31 @@ When Alan tries to compile this, he gets an error:
...
+Alan's code now compiles, but he also finds that his compile times have gone from under a second to around 6s, at least for a clean build.
+
+After Alan finishes adding the new trait, he shows his work off to Barbara and mentions he's happy with the work but is a little sad that compile times have worsened. Barbara, an experienced Rust developer, knows that using `async-trait` comes with some additional issues. In this particular case she is especially worried about tying their public API to a third-party dependency. Even though it is technically possible to implement traits annotated with `async_trait` without using `async_trait`, doing so in practice is very painful. For example `async_trait`:
+
+* handles lifetimes for you if the returned future is tied to the lifetime of some inputs.
+* boxes and pins the futures for you.
+
+which implementer will have to manually handle if they don't use `async_trait`. She decides to not worry Alan with this right now. Alan and Barbara are pretty happy with the results and go on to publish their crate which gets lots of users.
```
rylev, line 49:
```suggestion
which the implementer will have to manually handle if they don't use `async_trait`. She decides to not worry Alan with this right now. Alan and Barbara are pretty happy with the results and go on to publish their crate which gets lots of users.
```

src/vision/status_quo/alan_needs_async_in_traits.md
nikomatsakis: I don't know of any more!

zeenix:
> Once all resolved, I'd like to rebase to resolve the conflict and to squash all the `fixup` commits to the right commit.

Done.

PR #149, opened 2021-04-13
add story about bridging sync and async

nikomatsakis:
Cowritten with participants from writing workshop on 2021-04-02 at 14:30:

* Doc Jones
* Andrew Chin
* Emmanuel
* Frederik
* Zeeshan
* Trevor
* Ivan Tham

Closes https://github.com/rust-lang/wg-async-foundations/issues/54

### DIFF
--- a/src/vision/status_quo/barbara_bridges_sync_and_async.md
+++ b/src/vision/status_quo/barbara_bridges_sync_and_async.md
@@ -92,7 +92,7 @@ async fn main() {
 }
 ```
 
-Everything seems to work ok on her laptop, but when she pushes the code to production, it deadlocks immediately. "What's this?" she says. Confused, she runs the code on her laptop a few more times, but it seems to work fine.
+Everything seems to work ok on her laptop, but when she pushes the code to production, it deadlocks immediately. "What's this?" she says. Confused, she runs the code on her laptop a few more times, but it seems to work fine. (There's a faq explaining what's going on. -ed.)
 
 She decides to try debugging. She fires up a debugger but finds it is isn't really giving her useful information about what is stuck (she has [basically the same problems that Alan has](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/alan_tries_to_debug_a_hang.html)). [She wishes she could get insight into tokio's state.](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_wants_async_insights.html)
 
@@ -242,6 +242,18 @@ Later on, she wants to call `aggregate` from another binary. This one doesn't ha
 ### How would this story have played out differently for the other characters?
 
 I would expect it would work out fairly similarly, except that the type errors and things might well have been more challenging for people to figure out, assuming they aren't already familiar with Rust.
+
+### Why did Barbara only get deadlocks in production, and not on her laptop?
+
+This is because the production instance she was using had only a single core, but her laptop is a multicore machine. The actual cause of the deadlocks is that `block_on` basically "takes over" the tokio worker thread, and hence the tokio scheduler cannot run. If that `block_on` is blocked on another future that will have to execute, then some other thread must take over of completing that future. On Barbara's multicore machine, there were more threads available, so the system did not deadlock. But on the production instance, there was only a single thread. Barbara could have encountered deadlocks on her local machine as well if she had enough instances of `block_on` running at once. 
+
+### Could the runtime have prevented the deadlock?
+
+One way to resolve this problem would be to have a runtime that creates more threads as needed. This is what was proposed [in this blog post](https://async.rs/blog/stop-worrying-about-blocking-the-new-async-std-runtime/), for example.
+
+Adapting the number of worker threads has downsides. It requires knowing the right threshold for creating new threads (which is fundamentally unknowable). The result is that the runtime will sometimes observe that some thread seems to be taking a long time and create new threads *just before* that thread was about to finish. These new threads generate overhead and lower the overall performance. It also requires work stealing and other techniques that can lead to work running on mulitple cores and having less locality. Systems tuned for maximal performance tend to prefer a single thread per core for this reason.
+
+If some runtimes are adaptive, that may also lead to people writing libraries which block without caring. These libraries would then be a performance or deadlock hazard when used on a runtime that is not adaptive.
 
 ### Why didn't Barbara just use the sync API for reqwest?
 
--- a/src/vision/status_quo/barbara_bridges_sync_and_async.md
+++ b/src/vision/status_quo/barbara_bridges_sync_and_async.md
@@ -172,7 +172,7 @@ fn aggregate(urls: &[Url]) -> Vec<Data> {
 
 But this isn't really helping, as `spawn` still yields a future. She's getting the same errors.
 
-### Async all the way
+### Trying out `join_all`
 
 She remembers now that this whole drama started because she was converting her `main` function to be `async`. Maybe she doesn't have to bridge between sync and async? She starts digging around in the docs and finds `futures::join_all`. Using that, she can change `aggregate` to be an async function too:
 
@@ -286,6 +286,12 @@ reqwest does offer a synchronous API, but it's not enabled by default, you have 
 Regardless, the synchronous reqwest API is actually itself implemented using `block_on`: so Barbara would have ultimately hit the same issues. Further, not all crates offer synchronous APIs -- some offer only async APIs. In fact, these same issues are probably the sources of those panics that Barbara encountered in the past.
 
 In general, though, embedded sync within async or vice versa works "ok", once you know the right tricks. Where things become challenging is when you have a "sandwich", with async-sync-async. 
+
+### Do people mix `spawn_blocking` and `spawn` successfully in real code?
+
+Yes! Here is [some code from perf.rust-lang.org][egsb] doing exactly that. The catch is that it winds up giving you a future in the end, which didn't work for Barbara because her code is embedded within an iterator (and hence she can't make things async "all the way down").
+
+[egsb]: https://github.com/rust-lang/rustc-perf/blob/3651aa744ab2a22e1adf96a698851165086ad835/site/src/main.rs#L35-L36
 
 ### What are other ways people could experience similar problems mixing sync and async?
 
--- a/src/vision/status_quo/barbara_bridges_sync_and_async.md
+++ b/src/vision/status_quo/barbara_bridges_sync_and_async.md
@@ -98,7 +98,7 @@ She decides to try debugging. She fires up a debugger but finds it is isn't real
 
 Frustrated, she starts reading the tokio docs more closely and she realizes that `tokio` runtimes offer their own `block_on` method. "Maybe using tokio's `block_on` will help?" she thinks, "Worth a try, anyway." She changes the `aggregate` function to use tokio's `block_on`:
 
-```rust=
+```rust
 fn block_on<O>(f: impl Future<Output = O>) -> O {
     let rt  = tokio::runtime::Runtime::new().unwrap();
     rt.block_on(f)
@@ -120,20 +120,20 @@ The good news is that the deadlock is gone. The bad news is that now she is gett
 
 ```rust
 fn aggregate(urls: &[&str]) -> Vec<String> {
-    let handle = Handle::current();
+    let handle = tokio::runtime::Handle::current();
     urls.iter()
         .map(|url| handle.block_on(do_web_request(url)))
         .collect()
 }
 ```
 
-But it seems to give her the same panic. 
+But this also seems to panic in the same way.
 
 ### Trying out `spawn_blocking`
 
 Reading more into this problem, she realizes she is supposed to be using `spawn_blocking`. She tries replacing `block_on` with `tokio::task::spawn_blocking`:
 
-```rust=
+```rust
 fn aggregate(urls: &[Url]) -> Vec<Data> {
     urls
         .iter()
@@ -254,6 +254,30 @@ One way to resolve this problem would be to have a runtime that creates more thr
 Adapting the number of worker threads has downsides. It requires knowing the right threshold for creating new threads (which is fundamentally unknowable). The result is that the runtime will sometimes observe that some thread seems to be taking a long time and create new threads *just before* that thread was about to finish. These new threads generate overhead and lower the overall performance. It also requires work stealing and other techniques that can lead to work running on mulitple cores and having less locality. Systems tuned for maximal performance tend to prefer a single thread per core for this reason.
 
 If some runtimes are adaptive, that may also lead to people writing libraries which block without caring. These libraries would then be a performance or deadlock hazard when used on a runtime that is not adaptive.
+
+### Is there any way to have kept `aggregate` as a synchronous function?
+
+Yes, Barbara could have written something like this:
+
+```rust
+fn aggregate(urls: &[Url]) -> Vec<Data> {
+    let handle = Handle::current();
+
+    urls.iter()
+        .map(|url| handle.block_on(do_web_request(url)))
+        .collect()
+}
+
+#[tokio::main]
+async fn main() {
+    let data = task::spawn_blocking(move || aggregate(&[Url, Url]))
+        .await
+        .unwrap();
+    println!("done");
+}
+```
+
+This `aggregate` function can only safely be invoked from inside a tokio `spawn_blocking` call, however, since `Handle::current` will only work in that context. She could also have used the original futures variant of `block_on`, in that case, and things would also work.
 
 ### Why didn't Barbara just use the sync API for reqwest?
 
--- a/src/vision/status_quo/barbara_bridges_sync_and_async.md
+++ b/src/vision/status_quo/barbara_bridges_sync_and_async.md
@@ -227,12 +227,12 @@ Later on, she wants to call `aggregate` from another binary. This one doesn't ha
 
 ### What are the morals of the story?
 
-* Some projects don't care about max performance and just want things to work.
+* Some projects don't care about max performance and just want things to work once the program compiles.
     * They would probably be happy with sync but as the most popular libraries for web requests, databases, etc, offer async interfaces, they may still be using async code.
 * There are contexts where you can't easily add an `await`.
     * For example, inside of an iterator chain.
     * Big block of existing code.
-* Mixing sync and async code (`block_on`) can cause deadlocks that are really painful to diagnose.
+* Mixing sync and async code can cause deadlocks that are really painful to diagnose, particularly when you have an async-sync-async sandwich.
 
 ### Why did you choose Barbara to tell this story?
 
@@ -291,7 +291,7 @@ In general, though, embedded sync within async or vice versa works "ok", once yo
 
 * Using `std::Mutex` in async code.
 * Calling the blocking version of an asynchronous API.
-    * For example, `reqwest::blocking`, the synchronous `[zbus`](https://gitlab.freedesktop.org/dbus/zbus/-/blob/main/zbus/src/proxy.rs#L121) and [`rumqtt`](https://github.com/bytebeamio/rumqtt/blob/8de24cbc0484f459246251873aa6c80be8b6e85f/rumqttc/src/client.rs#L224) APIs.
+    * For example, `reqwest::blocking`, the synchronous [`zbus`](https://gitlab.freedesktop.org/dbus/zbus/-/blob/main/zbus/src/proxy.rs#L121) and [`rumqtt`](https://github.com/bytebeamio/rumqtt/blob/8de24cbc0484f459246251873aa6c80be8b6e85f/rumqttc/src/client.rs#L224) APIs.
     * These are commonly implemented by using some variant of `block_on` internally.
     * Therefore they can lead to panics or deadlocks depending on what async runtime they are built from and used with.
 
--- a/src/vision/status_quo/barbara_bridges_sync_and_async.md
+++ b/src/vision/status_quo/barbara_bridges_sync_and_async.md
@@ -1,0 +1,245 @@
+# 😱 Status quo stories: Barbara bridges sync and async in `perf.rust-lang.org`
+
+[How To Vision: Status Quo]: ../how_to_vision/status_quo.md
+[the raw source from this template]: https://raw.githubusercontent.com/rust-lang/wg-async-foundations/master/src/vision/status_quo/template.md
+[`status_quo`]: https://github.com/rust-lang/wg-async-foundations/tree/master/src/vision/status_quo
+[`SUMMARY.md`]: https://github.com/rust-lang/wg-async-foundations/blob/master/src/SUMMARY.md
+[open issues]: https://github.com/rust-lang/wg-async-foundations/issues?q=is%3Aopen+is%3Aissue+label%3Astatus-quo-story-ideas
+[open an issue of your own]: https://github.com/rust-lang/wg-async-foundations/issues/new?assignees=&labels=good+first+issue%2C+help+wanted%2C+status-quo-story-ideas&template=-status-quo--story-issue.md&title=
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+### Introducing `block_on`
+
+Barbara is working on the code for [perf.rust-lang.org](https://perf.rust-lang.org/) and she wants to do a web request to load various intermediate results. She has heard that the `reqwest` crate is quite nice, so she decides to give it a try. She writes up an async function that does her web request:
+
+```rust
+async fn do_web_request(url: &Url) -> Data {
+    ...
+}
+```
+
+She needs to apply this async function to a number of urls. She wants to use the iterator map function, like so:
+
+```rust
+async fn do_web_request(url: &Url) -> Data {...}
+
+fn aggregate(urls: &[Url]) -> Vec<Data> {
+    urls
+        .iter()
+        .map(|url| do_web_request(url))
+        .collect()
+}
+
+fn main() {
+    /* do stuff */
+    let data = aggregate();
+    /* do more stuff */
+}
+```
+
+Of course, since `do_web_request` is an async fn, she gets a type error from the compiler:
+
+```
+error[E0277]: a value of type `Vec<Data>` cannot be built from an iterator over elements of type `impl Future`
+  --> src/main.rs:11:14
+   |
+11 |             .collect();
+   |              ^^^^^^^ value of type `Vec<Data>` cannot be built from `std::iter::Iterator<Item=impl Future>`
+   |
+   = help: the trait `FromIterator<impl Future>` is not implemented for `Vec<Data>`
+```
+
+"Of course," she thinks, "I can't call an async function from a closure." She decides that since she is not overly concerned about performance, so she decides she'll just use a call to [`block_on` from the `futures` crate](https://docs.rs/futures/0.3.14/futures/executor/fn.block_on.html) and execute the function synchronously:
+
+```rust
+async fn do_web_request(url: &Url) -> Data {...}
+
+fn aggregate(urls: &[Url]) -> Vec<Data> {
+    urls
+        .iter()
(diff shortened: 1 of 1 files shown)


nikomatsakis:
cc @Mark-Simulacrum

nikomatsakis:
cc @doc-jones @eminence @emmanuelantony2000 @Frederik-Baetens (I think...)  @zeenix @pickfire

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+### And the cycle begins again
+
+Later on, she wants to call `aggregate` from another binary. This one doesn't have an `async main`. This context is deep inside of an iterator chain and was previously entirely synchronous. She realizes it would be a lot of work to change all the intervening stack frames to be `async fn`, rewrite the iterators into streams, etc. She decides to just call `block_on` again, even though it make her nervous.
+
+## 🤔 Frequently Asked Questions
+
+### What are the morals of the story?
+
+* Some projects don't care about max performance and just want things to work.
+    * They would probably be happy with sync but as the most popular libraries for web requests, databases, etc, offer async interfaces, they may still be using async code.
```
rylev, line 204: `reqwest` offers a sync API. Does Barbara not know about this? Or does she know but wishes to use async because she plans on moving her code to be async?

rylev (commented):
Looking good, but the story feels a bit thin at the end. I think it could use a bit more detail to drive the points home.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+    let data = aggregate();
+    /* do more stuff */
+}
+```
+
+The code compiles, and it seems to work.
+
+### Introducing `block_on`
+
+As Barbara works on play, she realizes that she needs to do more and more async operations. She decides to convert her synchronous `main` function into an `async main`. She's using tokio, so she is able to do this very conveniently with the `#[tokio::main]` decorator:
```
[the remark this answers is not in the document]
nikomatsakis: I do

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+            .map(|url| do_web_request(url))
+    ).await;
+}
+```
+
+Things are working again now, so she is happy.
+
+### Filtering
+
+Later on, she would like to apply a filter to the aggregation operation. She realizes that if she wants to use the fetched data when doing the filtering, she has to filter the vector after the join has completed; `join_all` doesn't have a way to put a filter into the iterator chain like she wants. She is annoyed, but performance isn't critical, so it's ok.
```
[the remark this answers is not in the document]
nikomatsakis: The problem is the filtering requires inspecting the data. If this were an iterator, she could do 

```rust
self.iter().map(|...|).filter(...).collect()
```

but now she is doing the equivalent of


```rust
self.iter().map(|...|).collect::<Vec<_>>().filter(...).collect()
```

src/vision/status_quo/barbara_bridges_sync_and_async.md
nikomatsakis: We actually discuss this below -- even if she did use the sync API, she'd eventually have the same issues. However, it's a good question why she didn't to begin with. I think in the "real story" she's already using a mix of sync and async in various places and the API itself wasn't reqwest but rather something related to databases which did not necessarily have a sync variant. Perhaps I should just change from reqwest to that.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+```rust
+#[tokio::main]
+async fn main() {
+    /* do stuff */
+    let data = aggregate();
+    /* do more stuff */
+}
+```
+
+Everything seems to work ok on her laptop, but when she pushes the code to production, it deadlocks immediately. "What's this?" she says. Confused, she runs the code on her laptop a few more times, but it seems to work fine.
```
nikomatsakis, line 93: I should maybe add a FAQ to explain this. What's going on here is that her laptop has multiple cores, so the problem with `block_on` doesn't show up. The production server has only a single core because it's a lightweight virtual machine.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+Of course! `spawn_blocking`, like `map`, just takes a regular closure, not an async closure. She's getting a bit frustrated now. "Well," she thinks, "I can use `spawn` to get into an async context!" So she adds a call to `spawn` inside the `spawn_blocking` closure:
+
+```rust
+fn aggregate(urls: &[Url]) {
+    let data: Vec<Data> =
+        urls
+            .iter()
+            .map(|url| tokio::task::spawn_blocking(move || {
```
nikomatsakis, line 164: I should also link to the code Mark showed me, which did exactly this (only it actually compiles, because it's not in an iterator)

src/vision/status_quo/barbara_bridges_sync_and_async.md
Mark-Simulacrum: FWIW my recollection - not sure how much this is a mix of different experiences - but the reqwest sync API at least from my POV wasn't used because (a) you need to enable the blocking feature, it's not on by default - so additional hurdle and (b) I had vague poor experience - I think documented elsewhere in these stories - with panics when mixing blocking (sync) with async APIs.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+error[E0277]: a value of type `Vec<Data>` cannot be built from an iterator over elements of type `tokio::task::JoinHandle<impl futures::Future>`
+  --> src/main.rs:22:14
+   |
+22 |             .collect();
+   |              ^^^^^^^ value of type `Vec<Data>` cannot be built from `std::iter::Iterator<Item=tokio::task::JoinHandle<impl futures::Future>>`
+   |
+   = help: the trait `FromIterator<tokio::task::JoinHandle<impl futures::Future>>` is not implemented for `Vec<Data>`
+```
+
+Of course! `spawn_blocking`, like `map`, just takes a regular closure, not an async closure. She's getting a bit frustrated now. "Well," she thinks, "I can use `spawn` to get into an async context!" So she adds a call to `spawn` inside the `spawn_blocking` closure:
```
zeenix, line 157:
```suggestion
Of course! `spawn_blocking`, like `map`, just takes a regular closure, not an async closure since they're not supported in Rust. She's getting a bit frustrated now. "Well," she thinks, "I can use `spawn` to get into an async context!" So she adds a call to `spawn` inside the `spawn_blocking` closure:
```

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+* There are contexts where you can't easily add an `await`.
+    * For example, inside of an iterator chain.
+    * Big block of existing code.
+* Mixing sync and async code (`block_on`) can cause deadlocks that are really painful to diagnose.
+
+
+### Why did you choose Barbara to tell this story?
+
+* Because Mark (who experienced most of it) is a very experienced Rust developer.
+* Because you could experience this story regardless of language background or being new to Rust.
```
zeenix, line 214: Just thinking out loud here but a bit unsure Barbara would not rethink their use of sync iterator at the first sign of trouble, especially at the point where they change `aggregator` to be an async function. Surely, they'd know they need async iterators (Streams) and very likely know of APIs like [`unfold`](https://docs.rs/futures/0.3.14/futures/stream/fn.unfold.html) that can be used to create streams out of futures. Streams would also solve her filtering requirement. Maybe Alan would fit better here?

zeenix (changes requested):

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+    let data = aggregate();
+    /* do more stuff */
+}
+```
+
+Everything seems to work ok on her laptop, but when she pushes the code to production, it deadlocks immediately. "What's this?" she says. Confused, she runs the code on her laptop a few more times, but it seems to work fine.
+
+She decides to try debugging. She fires up a debugger but finds it is isn't really giving her useful information about what is stuck (she has [basically the same problems that Alan has](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/alan_tries_to_debug_a_hang.html)). [She wishes she could get insight into tokio's state.](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_wants_async_insights.html)
+
+Frustrated, she starts reading the tokio docs more closely and she realizes that `tokio` runtimes offer their own `block_on` method. "What the hey," she thinks, "let's see if that works any better." She changes the `aggregate` function to use tokio's `block_on`:
```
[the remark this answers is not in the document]
nikomatsakis: I wanted to use what the "hey" because it amuses me :) maybe I'll add some more extreme goofy thing.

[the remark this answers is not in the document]
zeenix: oh. I never heard "what the hey" so got confused. :)

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+
+The good news is that the deadlock is gone. The bad news is that now she is getting a panic:
+
+> thread 'main' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.'
+
+"Well," she thinks, "I could use that `Handle` API to get the current runtime, maybe that will work?"
+
+```rust
+fn aggregate(urls: &[&str]) -> Vec<String> {
+    let handle = Handle::current();
```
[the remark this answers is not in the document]
pickfire: Well, the other parts didn't include `use ...` so I didn't include it here either.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+}
+```
+
+Everything seems to work ok on her laptop, but when she pushes the code to production, it deadlocks immediately. "What's this?" she says. Confused, she runs the code on her laptop a few more times, but it seems to work fine.
+
+She decides to try debugging. She fires up a debugger but finds it is isn't really giving her useful information about what is stuck (she has [basically the same problems that Alan has](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/alan_tries_to_debug_a_hang.html)). [She wishes she could get insight into tokio's state.](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_wants_async_insights.html)
+
+Frustrated, she starts reading the tokio docs more closely and she realizes that `tokio` runtimes offer their own `block_on` method. "What the hey," she thinks, "let's see if that works any better." She changes the `aggregate` function to use tokio's `block_on`:
+
+```rust=
```
pickfire, line 99:
```suggestion
```rust
```

src/vision/status_quo/barbara_bridges_sync_and_async.md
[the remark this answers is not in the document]
pickfire: I think hell is a bit rude. We should censor that.

src/vision/status_quo/barbara_bridges_sync_and_async.md
pickfire: > I should maybe add a FAQ to explain this. What's going on here is that her laptop has multiple cores, so the problem with block_on doesn't show up. The production server has only a single core because it's a lightweight virtual machine.

I experienced exactly the same thing. It can be verified on the local machine IIRC with `#[tokio::main(flavor = "multi_thread", worker_threads = 1)]`. I also asked in the channel why this isn't enabled by default in dev builds. If we enable something like this, it could mask other concurrency bugs that involves multiple workers.

src/vision/status_quo/barbara_bridges_sync_and_async.md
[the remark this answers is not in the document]
nikomatsakis: I'll tweak it to *something* a bit more familiar to non-native speakers :)

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+fn block_on<O>(f: impl Future<Output = O>) -> O {
+    let rt  = tokio::runtime::Runtime::new().unwrap();
```
pickfire, line 101:
```suggestion
    let rt = tokio::runtime::Runtime::new().unwrap();
```

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+        .map(|url| block_on(do_web_request(url)))
+        .collect()
+}
+```
+
+The good news is that the deadlock is gone. The bad news is that now she is getting a panic:
+
+> thread 'main' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.'
+
+"Well," she thinks, "I could use that `Handle` API to get the current runtime, maybe that will work?"
```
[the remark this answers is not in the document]
pickfire: What I was thinking at that point of time is to "re-use" the handle, I thought it will solve the blocking issue.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+}
+```
+
+But it seems to give her the same panic. 
+
+### Trying out `spawn_blocking`
+
+Reading more into this problem, she realizes she is supposed to be using `spawn_blocking`. She tries replacing `block_on` with `tokio::task::spawn_blocking`:
+
+```rust=
```
pickfire, line 134:
```suggestion
```rust
```

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+
+### And the cycle begins again
+
+Later on, she wants to call `aggregate` from another binary. This one doesn't have an `async main`. This context is deep inside of an iterator chain and was previously entirely synchronous. She realizes it would be a lot of work to change all the intervening stack frames to be `async fn`, rewrite the iterators into streams, etc. She decides to just call `block_on` again, even though it make her nervous.
+
+## 🤔 Frequently Asked Questions
+
+### What are the morals of the story?
+
+* Some projects don't care about max performance and just want things to work.
```
pickfire, line 203:
```suggestion
* Some projects don't care about max performance and just want things to work when it compiles.
```
Synchronous code usually just works when it compiles, but in async code you need to have an idea of the context, whether if the async code is called within a runtime or if you are spawning a runtime within a runtime, all these are not known at compile-time.

src/vision/status_quo/barbara_bridges_sync_and_async.md
pickfire: In my experience, not all the case you can control whether the code is sync or async. Some libraries will build a sync-async later and only accept a sync layer from you. I was doing tokio-rustls code, I have a runtime at the main code and I only get to build on top of a sync code. So it is `async <-> sync <-> async` and the only part I can change is the last part (to avoid trouble), one reason could be that there is no `async fn` for `impl` in stable rust yet. Sample of similar code I worked with (I found the docs for low-level async code not as well written or less examples as compared to the sync code in other parts of rust):

https://github.com/FlorianUekermann/rustls-acme/blob/a3f31ca9a79b6b031dd8ea3d68777ee838f98675/src/simple.rs#L28
https://github.com/FlorianUekermann/rustls-acme/blob/a3f31ca9a79b6b031dd8ea3d68777ee838f98675/src/resolver.rs#L240

But usually, if I only do one request I would probably avoid using async and just use reqwest sync API because it's way easier without anything to look out for, enabling the blocking feature is definitely easier than dealing with all kinds of async foot guns.

> I had vague poor experience - I think documented elsewhere in these stories - with panics when mixing blocking (sync) with async APIs.

Mixing sync with async is okay, same for async with sync. But if you want async within sync within async (hamburger) then the trouble starts.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+    * They would probably be happy with sync but as the most popular libraries for web requests, databases, etc, offer async interfaces, they may still be using async code.
+* There are contexts where you can't easily add an `await`.
+    * For example, inside of an iterator chain.
+    * Big block of existing code.
+* Mixing sync and async code (`block_on`) can cause deadlocks that are really painful to diagnose.
```
pickfire, line 208:
```suggestion
* Mixing sync and async code (`block_on` or `block_in_place`) can cause deadlocks that are really painful to diagnose.
```
`block_on` is not the only solution to glue sync and async code, although it is preferred, in some cases it cannot be used and only `block_in_place` could be used instead.

src/vision/status_quo/barbara_bridges_sync_and_async.md
pickfire: No, even Barbara who is an experienced rust developer but new to async can still trip here. `Stream`s are all new things in async, when I got into it I am also a bit confused at first, I was wondering if `Stream` is exactly the same as `Iterator` which one item is done before another or if the items processed can interleave (and indeed it can interleave), even when I read the `Stream` docs multiple times.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+
+
+### Why did you choose Barbara to tell this story?
+
+* Because Mark (who experienced most of it) is a very experienced Rust developer.
+* Because you could experience this story regardless of language background or being new to Rust.
+
+### How would this story have played out differently for the other characters?
+
+I would expect it would work out fairly similarly, except that the type errors and things might well have been more challenging for people to figure out, assuming they aren't already familiar with Rust.
```
pickfire, line 218:
```suggestion
I would expect it would work out fairly similarly, except that the type errors and things might well have been more challenging for people to figure out, assuming they aren't already familiar with async Rust.
```

Familiarity with sync rust does not help much when coming to async I believe.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+
+### What are other ways people could experience similar problems mixing sync and async?
+
+* Using `std::Mutex` in async code.
+* Calling the blocking version of an asynchronous API.
+    * For example, `reqwest::blocking`, the synchronous `[zbus`](https://gitlab.freedesktop.org/dbus/zbus/-/blob/main/zbus/src/proxy.rs#L121) and [`rumqtt`](https://github.com/bytebeamio/rumqtt/blob/8de24cbc0484f459246251873aa6c80be8b6e85f/rumqttc/src/client.rs#L224) APIs.
```
pickfire, line 224:
```suggestion
    * For example, `reqwest::blocking`, the synchronous [`zbus`](https://gitlab.freedesktop.org/dbus/zbus/-/blob/main/zbus/src/proxy.rs#L121) and [`rumqtt`](https://github.com/bytebeamio/rumqtt/blob/8de24cbc0484f459246251873aa6c80be8b6e85f/rumqttc/src/client.rs#L224) APIs.
```

src/vision/status_quo/barbara_bridges_sync_and_async.md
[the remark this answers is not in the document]
zeenix: sure, the thing is that we didn't specify where this `Handle` API is from and it's not a well-known one (e.g I didn't know/remember it and had to look it up).

src/vision/status_quo/barbara_bridges_sync_and_async.md
zeenix: > I was wondering if `Stream` is exactly the same as `Iterator`

Maybe you didn't go through **the** [async book](https://rust-lang.github.io/async-book/05_streams/01_chapter.html), which AFAIK we've been assuming our characters go through (especially Barbara) and it introduces streams as such from the very beginning:

> The Stream trait is similar to Future but can yield multiple values before completing, similar to the Iterator trait from the standard library

and

> Similar to synchronous Iterators, there are many different ways to iterate over and process the values in a Stream. There are combinator-style methods such as map, filter, and fold, and their early-exit-on-error cousins try_map, try_filter, and try_fold.

but as I said, I was just thinking out loud.

pickfire: Sorry, I didn't go through the async book (actually I did but the time I read it there isn't `Stream` in the book). But even in the characters set, Barbara wasn't said to go through async book, it just says

> She is using Rust for its feeling of productivity, and she expects Async Rust to continue in that tradition.

> She wants a design that feels like the rest of Rust.

> She loves Rust and she expects Async Rust to share its overall values.

So I believe Barbara may not go through async book given that she thinks it behaves similarly.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+            }))
+            .collect();
+}
+```
+
+But this isn't really helping, as `spawn` still yields a future. She's getting the same errors.
+
+### Async all the way
+
+She remembers now that this whole drama started because she was converting her `main` function to be `async`. Maybe she doesn't have to bridge between sync and async? She starts digging around in the docs and finds `futures::join_all`. Using that, she can change `aggregate` to be an async function too:
```
Darksonn, line 177: Note that `join_all` has quadratic time complexity.

src/vision/status_quo/barbara_bridges_sync_and_async.md
zeenix: > But even in the characters set, Barbara wasn't said to go through async book

Oh, I didn't mean to imply that it's a declared assumption but during the meetings, that's what I understood was the assumption. @nikomatsakis what do you say?

src/vision/status_quo/barbara_bridges_sync_and_async.md
Darksonn: But does `block_in_place` cause deadlocks that are really painful to diagnose? I don't think it does.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,245 @@
...
+async fn aggregate(urls: &[Url]) {
+    let data: Vec<Data> = futures::join_all(
+        urls
+            .iter()
+            .map(|url| do_web_request(url))
+    ).await;
+}
+```
+
+Things are working again now, so she is happy.
```
emmanuelantony2000, line 189: I wanted to leave one more final solution to which she might reach.

```rust
fn aggregate(urls: &[Url], handle: Handle) -> Vec<Data> {
    urls.iter()
        .map(|url| handle.block_on(do_web_request(url)))
        .collect()
}

#[tokio::main]
async fn main() {
    let handle = Handle::current();
    let data = task::spawn_blocking(move || aggregate(&[Url, Url], handle))
        .await
        .unwrap();
}
```

pickfire: In some cases, `aggregate` function may not be changed (happened to me). She may not have the luxury to pass the handle to the child function all the time if a third party crate was involved.

nikomatsakis: She could use `Handle::current` from inside `spawn_blocking`, I believe. Question is, would that actually work? What would happen if she did that? I guess I can try it myself...

nikomatsakis: [It seems to work fine](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bac20beedbe6eacc0a2a6d7a4d993e8b)

Darksonn: Yes, we guarantee that `Handle::current()` works fine in `spawn_blocking`.

emmanuelantony2000: I wanted to leave one more thing here, `Handle::block_on(...)` was introduced in tokio 1.4. And this solution is definitely possible, but to reach till here, might require folks to read the documentation thoroughly.

src/vision/status_quo/barbara_bridges_sync_and_async.md
pickfire: I don't think so. But in this suggestion I am discussing that `block_in_place` is another solution to mix sync and async code.

src/vision/status_quo/barbara_bridges_sync_and_async.md
nikomatsakis: I added a FAQ to attempt to answer this.

src/vision/status_quo/barbara_bridges_sync_and_async.md
nikomatsakis: I added a FAQ. Regarding your experience of figuring out where it was stuck, I would say that this is [Barbara wants async insights](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_wants_async_insights.html)

src/vision/status_quo/barbara_bridges_sync_and_async.md
nikomatsakis: Added a FAQ with this variant.

src/vision/status_quo/barbara_bridges_sync_and_async.md
nikomatsakis: Added a note.

src/vision/status_quo/barbara_bridges_sync_and_async.md
nikomatsakis: I am not assuming folks go through the async book, just that Barbara is experienced with Rust but new to async. I would hazard a guess that few people sit down to read the async book -- most people I've talked to don't know it exists!

src/vision/status_quo/barbara_bridges_sync_and_async.md
nikomatsakis: I'll just remove the reference, it's not important

nikomatsakis:
OK, I believe I have addressed all the feedback here!

nikomatsakis:
Thanks for all the thoughtful suggestions.

pickfire (approved):
Looks good to me.

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,332 @@
...
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+### Introducing `block_on`
```
zeenix, line 18: sorry but still feels misplaced. Let's remove it from here and place it.. (separate comment below).
```suggestion
```

src/vision/status_quo/barbara_bridges_sync_and_async.md
```diff
@@ -0,0 +1,332 @@
...
+error[E0277]: a value of type `Vec<Data>` cannot be built from an iterator over elements of type `impl Future`
+  --> src/main.rs:11:14
+   |
+11 |             .collect();
+   |              ^^^^^^^ value of type `Vec<Data>` cannot be built from `std::iter::Iterator<Item=impl Future>`
+   |
+   = help: the trait `FromIterator<impl Future>` is not implemented for `Vec<Data>`
+```
+
+"Of course," she thinks, "I can't call an async function from a closure." She decides that since she is not overly concerned about performance, so she decides she'll just use a call to [`block_on` from the `futures` crate](https://docs.rs/futures/0.3.14/futures/executor/fn.block_on.html) and execute the function synchronously:
```
zeenix, line 61:
```suggestion
"Of course," she thinks, "I can't call an async function from a closure." 

### Introducing `block_on`

She decides that since she is not overly concerned about performance, so she decides she'll just use a call to [`block_on` from the `futures` crate](https://docs.rs/futures/0.3.14/futures/executor/fn.block_on.html) and execute the function synchronously:
```

zeenix (changes requested):

Issue #65, opened 2021-03-19
when cancellation goes wrong


nikomatsakis:
* Brief summary: I've heard a lot of people discuss hazards related to task cancellation -- I think this is because any `await` point is effectively a cancellation point, and people don't always write "cancellation-safe" code. I'd love to hear more concrete examples of where this happens.
* Character: not sure!
* Key points or morals (if known): not sure!

[cast]: https://rust-lang.github.io/wg-async-foundations/vision/characters.html

nikomatsakis:
Part of this blog post feels (maybe) related:

https://kevinhoffman.medium.com/rust-async-and-the-terrible-horrible-no-good-very-bad-day-348ebc836274

>     Lesson III — If you drop (or allow to be dropped) the tokio runtime, any pending tasks are cancelled.

jbr:
AsyncDrop would likely help with some of the pain of unexpected/external drops, I think.

For intentionally canceling futures and streams, it would be nice if something like like https://crates.io/crates/stop-token (or https://crates.io/crates/stopper, slightly different implementation of the same idea) were standardized

nikomatsakis:
@jbr I think it would help to figure that out if we had as many concrete examples as we can. One thing I'd like to drill into a bit is the connection of cancellation and this, because we may not always be able to do async drop in cancellation.

seanmonstar:
Another relevant story, https://tomaka.medium.com/a-look-back-at-asynchronous-rust-d54d63934a1c:

> ```rust
> async fn read_send(file: &mut File, channel: &mut Sender<...>) {
>   loop {
>     let data = read_next(file).await;
>     let items = parse(&data);
>     for item in items {
>       channel.send(item).await;
>     }
>   }
> }
> ```

> Each await point in asynchronous code represents a moment when execution might be interrupted, and control given back to the user of this future. This user can, at their will, decide to drop the future at this point, stopping its execution altogether.

> If the user calls the read_send function, then polls it until it reaches the second await point (sending on a channel), then destroys the read_send future, all the local variables (data, items, and item) are silently dropped. By doing so, the user will have extracted data from file, but without sending this data on channel. This data is simply lost.

Matthias247:
I've written https://gist.github.com/Matthias247/ffc0f189742abf6aa41a226fe07398a8 2 years ago :-) 

> For intentionally canceling futures and streams, it would be really nice if something like https://crates.io/crates/stop-token were standardized

[...]

PR #153, opened 2021-04-16
Alan builds a cache

o0Ignition0o:
This story is built upon the [meeing notes](https://hackmd.io/VvCd4IOnQneLlVKl4sr8Dg?view) and a discussion that happened in #65.

I'd love to iterate on that and polish it, please let me know how to make it better :) 🙏

### DIFF
--- a/src/vision/status_quo/alan_builds_a_cache.md
+++ b/src/vision/status_quo/alan_builds_a_cache.md
@@ -11,12 +11,14 @@ If you would like to expand on this story, or adjust the answers to the FAQ, fee
 
 [Alan][] is working on an HTTP server. The server makes calls to some other service. The performance of the downstream service is somewhat poor, so Alan would like to implement some basic caching.
 
+[Alan]: ../characters/alan.html
+
 Alan writes up some code which does the caching:
 
 ```rust
 async fn get_response(&mut self, key: String) {
     // Try to get the response from cache
-    if let Some(cached_response) self.cache.get(key) {
+    if let Some(cached_response) = self.cache.get(key) {
         self.channel.send(cached_response).await;
         return;
     }
@@ -32,7 +34,7 @@ async fn get_response(&mut self, key: String) {
 
 Alan is happy with how things are working, but notices every once in a while the downstream service hangs. To prevent that, Alan implements a timeout.
 
-He remembers from the documentation for his favorite runtime that there is the `race` function which can kick off two futures and polls both until one completes. (similar to tokio's [select](https://docs.rs/tokio/1.5.0/tokio/macro.select.html) and async-std's [race](https://docs.rs/async-std/1.9.0/async_std/future/trait.Future.html#method.race) for example).
+He remembers from the documentation for his favorite runtime that there is the `race` function which can kick off two futures and polls both until one completes (similar to tokio's [select](https://docs.rs/tokio/1.5.0/tokio/macro.select.html) and async-std's [race](https://docs.rs/async-std/1.9.0/async_std/future/trait.Future.html#method.race) for example).
 
 
 ```rust
@@ -80,7 +82,7 @@ async fn get_response(&mut self, key: String) {
 * Futures can be "canceled" at any await point. Authors of futures must be aware that after an await, the code might not run.
     * This is similar to `panic` safety but way more likely to happen
 * Futures might be polled to completion causing the code to work. But then many years later, the code is changed and the future might conditionally not be polled to completion which breaks things.
-* The burden falls on the user of to poll to completion, and there is no way for the lib author to enforce this - they can only document this invariant.
+* The burden falls on the user of the future to poll to completion, and there is no way for the lib author to enforce this - they can only document this invariant.
 * Diagnosing and ultimately fixing this issue requires a fairly deep understanding of the semantics of futures.
 * Without a Barbara, it might be hard to even know where to start: No lints are available, Alan is left with a normal debugger and `println!`.
 
--- a/src/vision/status_quo/alan_builds_a_cache.md
+++ b/src/vision/status_quo/alan_builds_a_cache.md
@@ -73,6 +73,7 @@ async fn get_response(&mut self, key: String) {
 }
 ```
 
+## 🤔 Frequently Asked Questions
 
 ### **What are the morals of the story?**
 
@@ -82,3 +83,13 @@ async fn get_response(&mut self, key: String) {
 * The burden falls on the user of to poll to completion, and there is no way for the lib author to enforce this - they can only document this invariant.
 * Diagnosing and ultimately fixing this issue requires a fairly deep understanding of the semantics of futures.
 * Without a Barbara, it might be hard to even know where to start: No lints are available, Alan is left with a normal debugger and `println!`.
+
+### **What are the sources for this story?**
+The relevant sources of discussion for this story have been gathered [in this github issue](https://github.com/rust-lang/wg-async-foundations/issues/65).
+
+### **Why did you choose Alan to tell this story?**
+Alan has enough experience and understanding of push based async languages to make the assumptions that will trigger the bug.
+
+### **How would this story have played out differently for the other characters?**
+This story would likely have played out the same for almost everyone but Barbara, who has probably been bitten by that already.
+The debugging and fixing time would however probably have varied depending on experience and luck.
--- a/src/vision/status_quo/alan_builds_a_cache.md
+++ b/src/vision/status_quo/alan_builds_a_cache.md
@@ -1,0 +1,62 @@
+# 😱 Status quo stories: Alan tries to cache requests, which doesn't always happen.
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+
+## The story
+
+[Alan][] is working on an HTTP server. The server makes calls to some other service. The performance of the downstream service is somewhat poor, so Alan would like to implement some basic caching.
+
+Alan writes up some code which does the caching:
+
+```rust,ignore
+async fn get_response(&mut self, key: String) {
+    if let Some(cached_response) self.cache.get(key) {
+        self.channel.send(cached_response).await;
+        return;
+    }
+
+    let response = self.http_client.make_request(key).await;
+    
+    self.channel.send(response).await;
+    
+    self.cache.set(key, response);
+}
+```
+
+Alan is happy with how things are working, but notices every once in a while the downstream service hangs. To prevent that, Alan implements a timeout. 
+He remembers from the documentation for his favorite runtime that there is the `race` function which can kick off two futures and polls both until one completes:
+
+```rust ,ignore
+runtime::race(timeout(), get_response(key)).await
+```
+
+## The bug
+
+Alan ships to production but after several weeks he notices some users complaining that the receive old data. 
+
+Alan tries debugging. He uses his old friend `println!`. After hours of working through, he notices that sometimes the line that sets the response in the cache never gets called. 
+
+## The solution
+
+Alan goes to [Barbara][] and asks why in the world that might be ⁉️
+
+💡 Barbara looks through the code and notices that there is an await point between sending the response over the channel and setting the cache.
+
+This means the cache might not be set.
+
+Alan fixes it by setting the cache before sending the result over the channel. 🎉
+
+
+## Morals 
+
+* Futures can be "canceled" at any await point. Authors of futures must be aware that after an await, the code might not run. 
+* This is similar to `panic` safety but way more likely to happen 
+* Futures might be polled to completion causing the code to work. But then many years later, the code is changed and the future might conditionally not be polled to completion which breaks things. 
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+Alan is happy with how things are working, but notices every once in a while the downstream service hangs. To prevent that, Alan implements a timeout. 
+He remembers from the documentation for his favorite runtime that there is the `race` function which can kick off two futures and polls both until one completes:
+
+```rust ,ignore
+runtime::race(timeout(), get_response(key)).await
+```
+
+## The bug
+
+Alan ships to production but after several weeks he notices some users complaining that the receive old data.
```
Stupremee, line 40:
```suggestion
Alan ships to production but after several weeks he notices some users complaining that they receive old data. 
```

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+        self.channel.send(cached_response).await;
+        return;
+    }
+
+    let response = self.http_client.make_request(key).await;
+    
+    self.channel.send(response).await;
+    
+    self.cache.set(key, response);
+}
```
rylev, line 28: I'd add three code comments just to better describe what's happening:

```suggestion
async fn get_response(&mut self, key: String) {
    // Try to get the response from cache
    if let Some(cached_response) self.cache.get(key) {
        self.channel.send(cached_response).await;
        return;
    }
 
    // Get the response from the downstream service
    let response = self.http_client.make_request(key).await;
    self.channel.send(response).await;
    
    // Store the response in the cache
    self.cache.set(key, response);
}
```

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Alan is happy with how things are working, but notices every once in a while the downstream service hangs. To prevent that, Alan implements a timeout. 
+He remembers from the documentation for his favorite runtime that there is the `race` function which can kick off two futures and polls both until one completes:
```
rylev, line 32: Maybe we should add a note that this is similar to tokio's [`select`](https://docs.rs/tokio/1.5.0/tokio/macro.select.html) and async-std's [`race`](https://docs.rs/async-std/1.9.0/async_std/future/trait.Future.html#method.race).

src/vision/status_quo/alan_builds_a_cache.md
rylev, line 40:
```suggestion
Alan ships to production but after several weeks he notices some users complaining that they receive old data. 
```

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+
+Alan tries debugging. He uses his old friend `println!`. After hours of working through, he notices that sometimes the line that sets the response in the cache never gets called. 
+
+## The solution
+
+Alan goes to [Barbara][] and asks why in the world that might be ⁉️
+
+💡 Barbara looks through the code and notices that there is an await point between sending the response over the channel and setting the cache.
+
+This means the cache might not be set.
```
rylev, line 50: Perhaps explain why this might be - the cache might not be set because the future is dropped while the future is waiting for the response to be sent over the channel and thus the future is never polled again.

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+
+Alan fixes it by setting the cache before sending the result over the channel. 🎉
+
+
+## Morals
```
rylev, line 55: This section should be made to fit the same outline as the FAQ's found in other status quo stories [like here](https://github.com/rust-lang/wg-async-foundations/blob/master/src/vision/status_quo/alan_finds_database_drops_hard.md#-frequently-asked-questions)

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+
+This means the cache might not be set.
+
+Alan fixes it by setting the cache before sending the result over the channel. 🎉
+
+
+## Morals 
+
+* Futures can be "canceled" at any await point. Authors of futures must be aware that after an await, the code might not run. 
+* This is similar to `panic` safety but way more likely to happen
```
rylev, line 58: I believe this is a sub-point of the point above.

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+## Morals 
+
+* Futures can be "canceled" at any await point. Authors of futures must be aware that after an await, the code might not run. 
+* This is similar to `panic` safety but way more likely to happen 
+* Futures might be polled to completion causing the code to work. But then many years later, the code is changed and the future might conditionally not be polled to completion which breaks things. 
+* The burden falls on the user of to poll to completion, and there is no way for the lib author to enforce this - they can only document this invariant. 
+* Diagnosing and ultimately fixing this issue requires a fairly deep understanding of the semantics of futures. 
+* Without a Barbara, it might be hard to even know where to start.
```
rylev, line 62: Beyond knowing how futures work, there's also no tooling to help with this. You really only have a normal debugger or `println!` debugging to try to investigate.

rylev (changes requested):
Good start 🎉, but it still seems to be not fully moved to fit the structure that other status quo stories have.

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+Alan fixes it by setting the cache before sending the result over the channel. 🎉
+
+
+## Morals 
+
+* Futures can be "canceled" at any await point. Authors of futures must be aware that after an await, the code might not run. 
+* This is similar to `panic` safety but way more likely to happen 
+* Futures might be polled to completion causing the code to work. But then many years later, the code is changed and the future might conditionally not be polled to completion which breaks things. 
+* The burden falls on the user of to poll to completion, and there is no way for the lib author to enforce this - they can only document this invariant. 
+* Diagnosing and ultimately fixing this issue requires a fairly deep understanding of the semantics of futures.
```
Stupremee, line 61: We could also mention that there are no useful tools to debug such issues. Maybe even mention it in the Story?

Stupremee:
I forgot to submit my review and now Ryan mentioned the same points. Whoops

o0Ignition0o:
Thanks for the feedback, both of you 🙏 I'll try to address them this afternoon!

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,62 @@
...
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+
+## The story
+
+[Alan][] is working on an HTTP server. The server makes calls to some other service. The performance of the downstream service is somewhat poor, so Alan would like to implement some basic caching.
+
+Alan writes up some code which does the caching:
+
+```rust,ignore
```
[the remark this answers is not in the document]
pickfire:
```suggestion
```rust
```

o0Ignition0o:
I've adressed a couple of small nits for now. I'll try to work my way towards a relevant FAQ asap!

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,84 @@
...
+
+This means the cache might not be set.
+
+Alan fixes it by setting the cache before sending the result over the channel. 🎉
+
+```rust
+async fn get_response(&mut self, key: String) {
+    // ... cache miss happened here
+
+    // Yield while we wait for the http request to complete
```
o0Ignition0o, line 65: @rylev @Stupremee do you think using the `Yield` keyword is appropriate here? or is it wrong/misleading?

Stupremee: I feel like `yield` implies some sort of explicit loop or waiting process, which is not present here.
I think something simple like "Perform the HTTP request." should be fine

o0Ignition0o: given it's the new example, I'm trying to convey that both await points may or may not resume execution, which is why we wanna set the cache there, but yeah `yield` is confusing.

Any idea of a word that would allow us to convey the meaning better ?

Otherwise I'll just remove this example since its just two line swaps away from the first one so i guess it wouldn't be interesting anymore

Stupremee: "We perform the HTTP request and our code *might* continue after this `.await` once the HTTP request is complete"

Is probably a bit simpler to understand

o0Ignition0o: definitely! thanks!

o0Ignition0o:
I think I've adressed most of the comments, ready for the next pass! 😁

rylev (approved):
Awesome! 🎉

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,95 @@
...
+## The story
+
+[Alan][] is working on an HTTP server. The server makes calls to some other service. The performance of the downstream service is somewhat poor, so Alan would like to implement some basic caching.
+
+Alan writes up some code which does the caching:
+
+```rust
+async fn get_response(&mut self, key: String) {
+    // Try to get the response from cache
+    if let Some(cached_response) self.cache.get(key) {
```
nikomatsakis, line 19:
```suggestion
    if let Some(cached_response) = self.cache.get(key) {
```

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,95 @@
...
+    self.channel.send(response).await;
+    
+    // Store the response in the cache
+    self.cache.set(key, response);
+}
+```
+
+Alan is happy with how things are working, but notices every once in a while the downstream service hangs. To prevent that, Alan implements a timeout.
+
+He remembers from the documentation for his favorite runtime that there is the `race` function which can kick off two futures and polls both until one completes. (similar to tokio's [select](https://docs.rs/tokio/1.5.0/tokio/macro.select.html) and async-std's [race](https://docs.rs/async-std/1.9.0/async_std/future/trait.Future.html#method.race) for example).
```
nikomatsakis, line 35:
```suggestion
He remembers from the documentation for his favorite runtime that there is the `race` function which can kick off two futures and polls both until one completes (similar to tokio's [select](https://docs.rs/tokio/1.5.0/tokio/macro.select.html) and async-std's [race](https://docs.rs/async-std/1.9.0/async_std/future/trait.Future.html#method.race) for example).
```

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,95 @@
...
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+
+## The story
+
+[Alan][] is working on an HTTP server. The server makes calls to some other service. The performance of the downstream service is somewhat poor, so Alan would like to implement some basic caching.
```
zeenix, line 12: Missing link.

zeenix (approved):
Looks fine apart from the small issue.

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+## 🤔 Frequently Asked Questions
+
+### **What are the morals of the story?**
+
+* Futures can be "canceled" at any await point. Authors of futures must be aware that after an await, the code might not run.
+    * This is similar to `panic` safety but way more likely to happen
+* Futures might be polled to completion causing the code to work. But then many years later, the code is changed and the future might conditionally not be polled to completion which breaks things.
+* The burden falls on the user of to poll to completion, and there is no way for the lib author to enforce this - they can only document this invariant.
```
nikomatsakis, line 83:
```suggestion
* The burden falls on the user of the future to poll to completion, and there is no way for the lib author to enforce this - they can only document this invariant.
```

src/vision/status_quo/alan_builds_a_cache.md
```diff
@@ -0,0 +1,95 @@
...
+* Diagnosing and ultimately fixing this issue requires a fairly deep understanding of the semantics of futures.
+* Without a Barbara, it might be hard to even know where to start: No lints are available, Alan is left with a normal debugger and `println!`.
```
nikomatsakis, line 85: One thing not represented here, and that's fine, I'm not sure it should be, is that cancellation is *also* one of Rust's "super powers" somehow. Or so I've been told.

It might be worth noting that -- *unlike* panic safety -- there is a visual indicator of those points where cancellation may occur, which provides a plausible basis with which to do lints, and at least helps when auditing.

nikomatsakis (changes requested):
This is great. Left a few nits.

nikomatsakis:
I think this is basically ready to merge

src/vision/status_quo/alan_builds_a_cache.md
nikomatsakis:
```suggestion
[Alan][] is working on an HTTP server. The server makes calls to some other service. The performance of the downstream service is somewhat poor, so Alan would like to implement some basic caching.

[Alan]: ../characters/alan.html
```

Issue #155, opened 2021-04-16
Moving to a different executor


rylev:
### Brief summary

We have a few [stories](https://github.com/rust-lang/wg-async-foundations/blob/f583e8284e521238ce63b670d9371daf1262c1e0/src/vision/status_quo/barbara_anguishes_over_http.md) which highlight the fact that choosing a runtime upfront can be quite expensive. However, we don't yet have a story about a mature application which has been built around a certain executor, deciding to make the switch to another. 

This is because as things stand today, executors are usually bundled with runtimes that offer more than just a backing executor but also many convenience APIs that work only with that executor. Combined with the fact that most libraries are *not* runtime agnostic, this can really tie an application to a specific runtime. 

I think the story would likely take the form of a character having written a service for work using an off-the-shelf runtime like tokio. This works great for many years, but then as requirements change and the team behind the application has more and more specific needs, they decide that it would be best if they have their own executor. They then realize what a daunting task this would be as their entire application makes subtle and not-so-subtle assumptions that they are running on a specific executor. 

### Optional details

* (Optional) Which [character(s)] would be the best fit and why?
    * [ ] [Alan]: the experienced "GC'd language" developer, new to Rust
    * [ ] [Grace]: the systems programming expert, new to Rust
    * [ ] [Niklaus]: new programmer from an unconventional background
    * [x] [Barbara]: the experienced Rust developer
      * The problem here is not one that comes from making uninformed decisions. If Barbara had a crystal ball, she maybe could have done more work to make switching executors easier, but tokio was a really great choice for a very long time. Unforeseen changes in requirements have caused the building need to switch to a bespoke executor and the ecosystem makes this hard to actually do.   
*  What are the key points or morals to emphasize?
    * Switching executors is hard because executors are almost always bundled with rich runtimes that provide convenient APIs that ties applications to the runtime.  

[character(s)]: https://rust-lang.github.io/wg-async-foundations/vision/characters.html
[project(s)]: https://rust-lang.github.io/wg-async-foundations/vision/projects.html
[Alan]: https://rust-lang.github.io/wg-async-foundations/vision/characters/alan.html
[Grace]: https://rust-lang.github.io/wg-async-foundations/vision/characters/grace.html
[Niklaus]: https://rust-lang.github.io/wg-async-foundations/vision/characters/niklaus.html
[Barbara]: https://rust-lang.github.io/wg-async-foundations/vision/characters/barbara.html

PR #160, opened 2021-04-17
[shiny future] Alan's trust in the compiler is rewarded, Alan switches runtimes

nikomatsakis:
A pair of shiny future stories sketching out an alternative path for Alan as he gets started, and also some ideas for how he might switch between runtimes.

This is partly based on the writing session we had yesterday, so co-author credit goes to @Mark-Simulacrum, @tmandry, @emmanuelantony2000, and @doc-jones.

### DIFF
--- a/src/vision/shiny_future/alan_switches_runtimes.md
+++ b/src/vision/shiny_future/alan_switches_runtimes.md
@@ -83,7 +83,7 @@ Good question! I'm not entirely sure! I have to go looking and think about it. M
 * There is some way to seamlessly change to a different default runtime to use for `async fn main`.
 * There is no global runtime, just the current runtime.
 * When you are using this different runtime, you can write code that is hard-coded to it and which exposes additional capabilities.
-* You can integrate multiple runtimes relatively easily, and the std APIs work with whechever is the current runtime.
+* You can integrate multiple runtimes relatively easily, and the std APIs work with whichever is the current runtime.
 
 ### How do you imagine the std APIs and so forth know the current runtime?
 
--- a/src/vision/shiny_future/alan_switches_runtimes.md
+++ b/src/vision/shiny_future/alan_switches_runtimes.md
@@ -26,8 +26,6 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 He asks Barbara, one of his coworkers, "What is this `async_runtime` annotation? And what's `humboldt`?" She answers by explaining to him that Rust's support for async I/O is actually based around an underlying runtime. "Rust gives you a pretty decent runtime by default," she says, "but it's not tuned for our workloads. We wrote our own runtime, which we call `humboldt`."
 
-[DistriData]: ../projects/DistriData.md
-
 Alan asks, "What happens with the various std APIs? For example, I see we are calling `std::async_thread::spawn` -- when I used that before, it spawned tasks into the default runtime. What happens now?"
 
 Barbara explains that the "async" APIs in std generally execute relative to the current runtime that is in use. "When you call `std::async_thread::spawn`, it will spawn a task onto the current runtime. It's the same with the routines in `std::async_io` and and so forth. The `async_runtime` annotation just affects which runtime runs the `main` function, everything else follows from there."
@@ -117,7 +115,6 @@ Projects like [DistriData] really benefit from being able to customize their run
 
 We have to pay careful attention to embedded projects like [MonsterMesh]. Some of the most obvious ways to implement this future would lean on `dyn` types and perhaps boxing, and that would rule out embedded projects. Embedded runtimes like [embassy] are also the most different in their overall design and they would have the hardest time fitting into the std APIs (of course, many embedded projects are already no-std, but many of them make use of some subset of the std capabilities through the facade).
 
-[MonsterMesh]: ../projects/MonsterMesh.md
 [embassy]: https://github.com/akiles/embassy
 
 ### **What are the incremental steps towards realizing this shiny future?**
@@ -134,7 +131,6 @@ There are a few steps required to realize this future:
 
 Yes. We will need to collaborate to define traits that std can use to interface with each runtime, and the runtimes will need to implement those traits. This is going to be non-trivial, because we want to preserve the ability for independent runtimes to experiment, while also preserving the ability to "max and match" and re-use components. For example, it'd probably be useful to have a bunch of shared I/O infrastructure, or to have utility crates for locks, for running threadpools, and the like. On the other hand, tokio takes advantage of the fact that it owns the I/O types *and* the locks *and* the scheduler to do some [nifty tricks](https://tokio.rs/blog/2020-04-preemption) and we would want to ensure that remains an option.
 
-### Is this story 
 
 [character]: ../characters.md
 [comment]: ./comment.md
@@ -146,3 +142,5 @@ Yes. We will need to collaborate to define traits that std can use to interface 
 [projects]: ../projects.md
 [htvsq]: ../how_to_vision/shiny_future.md
 [cannot be wrong]: ../how_to_vision/comment.md#comment-to-understand-or-improve-not-to-negate-or-dissuade
+[DistriData]: ../projects/DistriData.md
+[MonsterMesh]: ../projects/MonsterMesh.md
--- a/src/vision/shiny_future/alan_switches_runtimes.md
+++ b/src/vision/shiny_future/alan_switches_runtimes.md
@@ -28,7 +28,7 @@ He asks Barbara, one of his coworkers, "What is this `async_runtime` annotation?
 
 Alan asks, "What happens with the various std APIs? For example, I see we are calling `std::async_thread::spawn` -- when I used that before, it spawned tasks into the default runtime. What happens now?"
 
-Barbara explains that the "async" APIs in std generally execute relative to the current runtime that is in use. "When you call `std::async_thread::spawn`, it will spawn a task onto the current runtime. It's the same with the routines in `std::async_io` and and so forth. The `async_runtime` annotation just affects which runtime runs the `main` function, everything else follows from there."
+Barbara explains that the "async" APIs in std generally execute relative to the current runtime that is in use. "When you call `std::async_thread::spawn`, it will spawn a task onto the current runtime. It's the same with the routines in `std::async_io` and so forth. The `async_runtime` annotation just affects which runtime runs the `main` function, everything else follows from there."
 
 ## Learning more about Humboldt
 
@@ -38,7 +38,7 @@ Alan sees that some of the networking code that is being used in their applicati
 use humboldt::network;
 ```
 
-He asks Barbara, "Why don't we use the `std::async_io` APIs for that?" She explains that Humbolt makes use of some custom kernel extensions that, naturally enough, aren't part of the std library. "TCP is for rubes," she says, "we are using TTCP -- Turbo TCP." Her mind wanders briefly to [Turbo Pascal] and she has a brief moment of yearning for the days when computers had a "Turbo" button that changed them from 8 MHz to 12 MHz. She snaps back into the present day. "Anyway, the `std::async_io` APIs just call into humboldt's APIs via various traits. But we can code directly against `humboldt` when we want to access the extra capabilities it offers."
+He asks Barbara, "Why don't we use the `std::async_io` APIs for that?" She explains that Humboldt makes use of some custom kernel extensions that, naturally enough, aren't part of the std library. "TCP is for rubes," she says, "we are using TTCP -- Turbo TCP." Her mind wanders briefly to [Turbo Pascal] and she has a brief moment of yearning for the days when computers had a "Turbo" button that changed them from 8 MHz to 12 MHz. She snaps back into the present day. "Anyway, the `std::async_io` APIs just call into humboldt's APIs via various traits. But we can code directly against `humboldt` when we want to access the extra capabilities it offers."
 
 [Turbo Pascal]: https://en.wikipedia.org/wiki/Turbo_Pascal
 
@@ -76,7 +76,7 @@ Good question! I'm not entirely sure! I have to go looking and think about it. M
 
 ### What are the key points you were trying to convey with this status quo story?
 
-* There is some way to seamlessly change to a different default runtime to use for `async main`.
+* There is some way to seamlessly change to a different default runtime to use for `async fn main`.
 * There is no global runtime, just the current runtime.
 * When you are using this different runtime, you can write code that is hard-coded to it and which exposes additional capabilities.
 * You can integrate multiple runtimes relatively easily, and the std APIs work with whechever is the current runtime.
--- a/src/vision/shiny_future/alan_switches_runtimes.md
+++ b/src/vision/shiny_future/alan_switches_runtimes.md
@@ -1,4 +1,4 @@
-# ✨ Shiny future stories: template
+# ✨ Shiny future stories: Alan switches runtimes
 
 [How To Vision: Shiny Future]: ../how_to_vision/shiny_future.md
 [the raw source from this template]: https://raw.githubusercontent.com/rust-lang/wg-async-foundations/master/src/vision/shiny_future/template.md
--- a/src/vision/shiny_future/alan_switches_runtimes.md
+++ b/src/vision/shiny_future/alan_switches_runtimes.md
@@ -16,7 +16,7 @@ If you would like to expand on this story, or adjust the answers to the FAQ, fee
 Since his [early adventures](./alans_trust_in_the_compiler_is_rewarded.md) with Async I/O went so well, Alan has been looking for a way to learn more. He finds a job working in Rust. One of the projects he works on is [DistriData]. Looking at their code, he sees an annotation he has never seen before:
 
 ```rust
-#[async_runtime(humboldt)]
+#[humboldt::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let result = std::async_thread::spawn(async move {
         do_something()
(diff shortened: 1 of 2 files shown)


src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let result = std::async_thread::spawn(async move {
+        do_something()
+    });
+}
+```
+
+He asks Barbara, one of his coworkers, "What is this `async_runtime` annotation? And what's `humboldt`?" She answers by explaining to him that Rust's support for async I/O is actually based around an underlying runtime. "Rust gives you a pretty decent runtime by default," she says, "but it's not tuned for our workloads. We wrote our own runtime, which we call `humboldt`."
+
+[DistriData]: ../projects/DistriData.md
```
Stupremee, line 29: Why is the link here and not at the bottom of the page?

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+    * Depending on the details, language changes may be required. 
+* We have to hammer out the set of traits or other interfaces used to define the parts of a runtime (see below for some of the considerations).
+    * We can start with easier cases and proceed to more difficult ones, however.
+
+### **Does realizing this future require cooperation between many projects?**
+
+Yes. We will need to collaborate to define traits that std can use to interface with each runtime, and the runtimes will need to implement those traits. This is going to be non-trivial, because we want to preserve the ability for independent runtimes to experiment, while also preserving the ability to "max and match" and re-use components. For example, it'd probably be useful to have a bunch of shared I/O infrastructure, or to have utility crates for locks, for running threadpools, and the like. On the other hand, tokio takes advantage of the fact that it owns the I/O types *and* the locks *and* the scheduler to do some [nifty tricks](https://tokio.rs/blog/2020-04-preemption) and we would want to ensure that remains an option.
+
+### Is this story
```
Stupremee, line 137: Is this a leftover or a thing to do?

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+}
+```
+
+He asks Barbara, one of his coworkers, "What is this `async_runtime` annotation? And what's `humboldt`?" She answers by explaining to him that Rust's support for async I/O is actually based around an underlying runtime. "Rust gives you a pretty decent runtime by default," she says, "but it's not tuned for our workloads. We wrote our own runtime, which we call `humboldt`."
+
+[DistriData]: ../projects/DistriData.md
+
+Alan asks, "What happens with the various std APIs? For example, I see we are calling `std::async_thread::spawn` -- when I used that before, it spawned tasks into the default runtime. What happens now?"
+
+Barbara explains that the "async" APIs in std generally execute relative to the current runtime that is in use. "When you call `std::async_thread::spawn`, it will spawn a task onto the current runtime. It's the same with the routines in `std::async_io` and and so forth. The `async_runtime` annotation just affects which runtime runs the `main` function, everything else follows from there."
```
Stupremee, line 33:
```suggestion
Barbara explains that the "async" APIs in std generally execute relative to the current runtime that is in use. "When you call `std::async_thread::spawn`, it will spawn a task onto the current runtime. It's the same with the routines in `std::async_io` and so forth. The `async_runtime` annotation just affects which runtime runs the `main` function, everything else follows from there."
```

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Learning more about Humboldt
+
+Alan sees that some of the networking code that is being used in their application is creating network connections using humboldt APIs:
+
+```rust
+use humboldt::network;
+```
+
+He asks Barbara, "Why don't we use the `std::async_io` APIs for that?" She explains that Humbolt makes use of some custom kernel extensions that, naturally enough, aren't part of the std library. "TCP is for rubes," she says, "we are using TTCP -- Turbo TCP." Her mind wanders briefly to [Turbo Pascal] and she has a brief moment of yearning for the days when computers had a "Turbo" button that changed them from 8 MHz to 12 MHz. She snaps back into the present day. "Anyway, the `std::async_io` APIs just call into humboldt's APIs via various traits. But we can code directly against `humboldt` when we want to access the extra capabilities it offers."
```
Stupremee, line 43:
```suggestion
He asks Barbara, "Why don't we use the `std::async_io` APIs for that?" She explains that Humboldt makes use of some custom kernel extensions that, naturally enough, aren't part of the std library. "TCP is for rubes," she says, "we are using TTCP -- Turbo TCP." Her mind wanders briefly to [Turbo Pascal] and she has a brief moment of yearning for the days when computers had a "Turbo" button that changed them from 8 MHz to 12 MHz. She snaps back into the present day. "Anyway, the `std::async_io` APIs just call into humboldt's APIs via various traits. But we can code directly against `humboldt` when we want to access the extra capabilities it offers."
```

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+
+[Turbo Pascal]: https://en.wikipedia.org/wiki/Turbo_Pascal
+
+## Integrating into other event loops
+
+Later on, Alan is working on a visualizer front-end that integrates with [DistriData] to give more details about their workloads. To do it, he needs to integrate with Cocoa APIs and he wants to run certain tasks on Grand Central Dispatch. He approaches Barbara and asks, "If everything is running on `humboldt`, is there a way for me to run some things on another event loop? How does that work?"
```
Stupremee, line 49: What are the Cocoa APIs and what exactly is Grand central Dispatch?

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+
+Barbara explains, "That's easy. You just have to use the `gcd` wrapper crate -- you can find it on `crates.io`. It implements the runtime traits for `gcd` and it has a `spawn` method. Once you spawn your task onto `gcd`, everything you run within gdb will be running in that context."
```
Stupremee, line 51: What does the last part "everything you run within gdb" mean? GDB as the debugger?

src/vision/shiny_future/alan_switches_runtimes.md
Stupremee: From the following text I suspect that Grand Central Dispatch (gcd) is another runtime, is that right?

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+
+## 🤔 Frequently Asked Questions
+
+### What status quo story or stories are you retelling?
+
+Good question! I'm not entirely sure! I have to go looking and think about it. Maybe we'll have to write some more.
+
+### What are the key points you were trying to convey with this status quo story?
+
+* There is some way to seamlessly change to a different default runtime to use for `async main`.
```
Stupremee, line 81:
```suggestion
* There is some way to seamlessly change to a different default runtime to use for `async fn main`.
```

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+
+That feels like a portability question. See the (yet to be written) sequel story, "Alan runs some things on WebAssembly". =)
+
+### **What is [Alan] most excited about in this future? Is he disappointed by anything?**
+
+Alan is excited about how easy it is to get async programs up and running, and he finds that they perform pretty well once he does so, so he's happy.
+
+### **What is [Grace] most excited about in this future? Is she disappointed by anything?**
+
+Grace is concerned with memory safety and being able to deploy her tricks she knows from other languages. Memory safety works fine here. In terms of tricks she knows and loves, she's happy that she can easily switch to another runtime. The default runtime is good and works well for most things, but for the [`DistriData`] project, they really need something tailored just for them. She is also happy she can use the extended APIs offered by `humboldt`.
```
Stupremee, line 100: Grace would probably be concerned the overhead that is produced by all the traits, and stuff which might be important n certain embedded environments. Similarly to Barbara but specific to embedded environments.

Stupremee (commented):
I only read one story but it was a great read 👍🎉

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: typo, I meant "gcd". My fingers are very used to typing gdb, though.

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Integrating into other event loops
+
+Later on, Alan is working on a visualizer front-end that integrates with [DistriData] to give more details about their workloads. To do it, he needs to integrate with Cocoa APIs and he wants to run certain tasks on Grand Central Dispatch. He approaches Barbara and asks, "If everything is running on `humboldt`, is there a way for me to run some things on another event loop? How does that work?"
+
+Barbara explains, "That's easy. You just have to use the `gcd` wrapper crate -- you can find it on `crates.io`. It implements the runtime traits for `gcd` and it has a `spawn` method. Once you spawn your task onto `gcd`, everything you run within gdb will be running in that context."
+
+Alan says, "And so, if I want to get things running on `humboldt` again, I spawn a task back on `humboldt`?"
+
+"Exactly," says Barbara. "Humboldt has a global event loop, so you can do that by just doing `humboldt::spawn`. You can also just use the `humboldt::io` APIs directly and so forth if you want to do I/O with humboldt."
```
Mark-Simulacrum, line 55: I think it might be worth digging in a bit here - do the humboldt::io types always use the humboldt runtime? Or are they like std, where they use the 'current' runtime?

I think both are likely valid options, though the first of these may end up slightly faster...

Stupremee:
Why does this PR include the sync-async bridges story?

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,148 @@
...
+This is a draft "shiny future" story submitted as part of the brainstorming period. It is derived from what actual Rust users wish async Rust should be, and is meant to deal with some of the challenges that Async Rust programmers face today.
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as peoples needs and desires for async Rust may differ greatly, shiny future stories [cannot be wrong]. At worst they are only useful for a small set of people or their problems might be better solved with alternative solutions). Alternatively, you may wish to [add your own shiny vision story][htvsq]!
+
+## The story
+
+Since his [early adventures](./alans_trust_in_the_compiler_is_rewarded.md) with Async I/O went so well, Alan has been looking for a way to learn more. He finds a job working in Rust. One of the projects he works on is [DistriData]. Looking at their code, he sees an annotation he has never seen before:
+
+```rust
+#[async_runtime(humboldt)]
```
pickfire, line 19: Some async runtime runs on single thread and some runs on multiple threads. How do we different among it if we do it this way? Also, will the internal code changes along they way? If we force everything to be Arc-wrapped wouldn't we impose an extra cost on those using it on a single thread?

nikomatsakis: I don't quite follow. I am imagining something like this:

* If one writes `async fn main`, then the compiler generates a synchronous `main` function that invokes a method similar to `block_on` from some crate. You can specify the crate via the `async_runtime` annotation (or perhaps some other annotation; e.g., perhaps just the `#[humboldt::main]` convention that seems to be in wide use today).
* If one prefers, you can just write `fn main` yourself, which would allow you to instantiate runtimes with whatever options you like.

I can spell out some of this in a FAQ -- but I want to be sure first, does that address your point? It seems to me like it does. The runtime could indeed be single threaded or not, and have all kinds of other options besides.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: In what way do traits introduce overhead? Do you mean because of dynamic dispatch, or do you mean because having to "fit oneself" to a trait may induce overhead?

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: I would expect the former. I can clarify.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: [Grand Central Dispatch on wikipedia](https://en.wikipedia.org/wiki/Grand_Central_Dispatch) -- I don't really know whether it's the best example, I just wanted something that is specific to some setting.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: Editing error, I think :)

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: No particular reason :) I moved it down

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: I'm going to change to `humboldt::main`

src/vision/shiny_future/alan_switches_runtimes.md
Stupremee: Ah okay. No it's totally fine to use that, but a short description what it is + link to Wikipedia would help to understand it.

src/vision/shiny_future/alan_switches_runtimes.md
Stupremee: I assume that going through trait objects and probably even some generics, will generate much more code which could lead to exceeding some size limit on embedded devices. It may not even be true in reality but I thought it could be.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: Pushed some changes, @pickfire.

pickfire: No, that's not what I am talking about. I mean for example, `tokio::main` two different flavors `flavor = "multi-thread"` and `flavor = "current_thread"`, and also options for those flavors. How do we intend to support those with `#[async_runtime(...)]`? https://docs.rs/tokio/1.4.0/tokio/attr.main.html

Also, it is also common to see some runtime only supporting multi-thread such as async-std but others only supporting single-thread such as actix, and some such as tokio supports both. Is it possible to know how many threads it supports as in having an explicit information for this? And how can we pass in the additional information for tokio such as `worker_threads`?

But yeah, changing that to `#[humboldt::main]` seemed to fix the issue. Just that now the user needs to read the docs to know whether it is running with multi-thread or single-thread (like how it is done currently).

The other parts I mentioned can just ignore, it's probably not for this issue.

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,150 @@
...
+    gcd::spawn(async move {
+        let foo = do_something_on_gcd();
+
+        let bar = humboldt::spawn(async move {
+            do_a_little_bit_of_stuff_on_humboldt();
+        });
+
+        combine(foo.await, bar.await);
+    });
+}
```
pickfire, line 72: Shouldn't `humboldt::spawn` be out of `gcd::spawn`?

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,150 @@
...
+### What status quo story or stories are you retelling?
+
+Good question! I'm not entirely sure! I have to go looking and think about it. Maybe we'll have to write some more.
+
+### What are the key points you were trying to convey with this status quo story?
+
+* There is some way to seamlessly change to a different default runtime to use for `async fn main`.
+* There is no global runtime, just the current runtime.
+* When you are using this different runtime, you can write code that is hard-coded to it and which exposes additional capabilities.
+* You can integrate multiple runtimes relatively easily, and the std APIs work with whechever is the current runtime.
```
rylev, line 86:
```suggestion
* You can integrate multiple runtimes relatively easily, and the std APIs work with whichever is the current runtime.
```

src/vision/shiny_future/alan_switches_runtimes.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Learning more about Humboldt
+
+Alan sees that some of the networking code that is being used in their application is creating network connections using humboldt APIs:
+
+```rust
+use humboldt::network;
+```
+
+He asks Barbara, "Why don't we use the `std::async_io` APIs for that?" She explains that Humboldt makes use of some custom kernel extensions that, naturally enough, aren't part of the std library. "TCP is for rubes," she says, "we are using TTCP -- Turbo TCP." Her mind wanders briefly to [Turbo Pascal] and she has a brief moment of yearning for the days when computers had a "Turbo" button that changed them from 8 MHz to 12 MHz. She snaps back into the present day. "Anyway, the `std::async_io` APIs just call into humboldt's APIs via various traits. But we can code directly against `humboldt` when we want to access the extra capabilities it offers."
```
rylev, line 43: > But we can code directly against `humboldt` when we want to access the extra capabilities it offers.

I'd like to see how this interacts with the ability to easily switch runtimes (which is outlined in https://github.com/rust-lang/wg-async-foundations/issues/155)

rylev (commented):
This is looking really good! I don't have much to add.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: I wonder if a more realistic example would be to have multiple thread-pools, with e.g. different priorities or something. I'm going to leave it how it is, regardless.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: Yeah, that's partly why I changed to `humboldt::main`, it gives more control to the crate to define exactly what options. I definitely agree that the user will have to read the executor to understand what it means to use it. I think that's fine. A more interesting question is how that affects the "interoperability" traits, which have to be targeted for maximum flexibility, but the point of the "shiny future" is not to have all the answers.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: I don't know that it's true, but it is plausible. I did add some notes about embedded devices in the FAQ -- in terms of what sorts of projects would be affected -- I can add some comments to this effect as well.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: It seems clear that the more you really directly on the APIs specific to one runtime, the harder it will be to change later. I could add a note about this, I'm not sure what else to sketch though. (Thoughts?)

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: Done.

src/vision/shiny_future/alan_switches_runtimes.md
nikomatsakis: Added a comment. Going to mark as done for now, but feel free to leave a fresh comment.

PR #165, opened 2021-04-21
[shiny-future] recursive async fns with auto-box

nikomatsakis:
A story about how we could enable recursive functions if we wanted to.

My goals with this story was to paint a version where recursive async functions just worked with no annotations on the part of the user, which implies some amount of automatic allocation on the part of the compiler (though far less than you would find in other languages, where every async function call implies an allocation).

### DIFF
--- a/src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
+++ b/src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
@@ -1,0 +1,119 @@
+# ✨ Shiny future stories: Grace writes a recursive function with auto-box
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "shiny future" story submitted as part of the brainstorming period. It is derived from what actual Rust users wish async Rust should be, and is meant to deal with some of the challenges that Async Rust programmers face today.
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as peoples needs and desires for async Rust may differ greatly, shiny future stories [cannot be wrong]. At worst they are only useful for a small set of people or their problems might be better solved with alternative solutions). Alternatively, you may wish to [add your own shiny vision story][htvsq]!
+
+## The story
+
+### Early days
+
+Grace is using async Rust for the first time. She is experimenting with async functions and she begins with her favorite "Hello, World" sort of application, Fibonacci:
+
+```rust
+async fn fibonacci(n: u32) -> u32 {
+    match n {
+        0 => 1,
+        1 => 1,
+        _ => fibonacci(n - 1) + fibonacci(n - 2),
+    }
+}
+```
+
+It doesn't make much sense, but that's ok. She compiles it as part of a webapp, so that when she requests `https://localhost/fib/22` she gets back a page with the correct result. Fun!
+
+### Discovering automatic boxing
+
+Now that Grace has the basics working, she starts adding code to the [MonsterMesh] project that she is developing at work. She is writing a simple recursive function, similar to `fibonacci`, but she finds that she gets an error:
+
+```
+error: `Box` required for async function
+ --> src/lib.rs:4:9
+  |
+4 |     async fn fibonacci(x: usize) -> usize {
+  |              ^^^^^^^^^ the function `fibonacci` is recursive and requires a `Box`
+  |
+note: the lint level is defined here
+ --> src/lib.rs:1:9
+  |
+1 | #![deny(automatic_box)]
+  |         ^^^^^^^^^^^^^
+help: to request a box explicitly, write `box async fn`
+  |
+4 |     box async fn fibonacci(x: usize) -> usize {
+```
+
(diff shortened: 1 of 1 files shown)


src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
```diff
@@ -0,0 +1,119 @@
...
+### Early days
+
+Grace is using async Rust for the first time. She is experimenting with async functions and she begins with her favorite "Hello, World" sort of application, Fibonacci:
+
+```rust
+async fn fibonacci(n: u32) -> u32 {
+    match n {
+        0 => 1,
+        1 => 1,
+        _ => fibonacci(n - 1) + fibonacci(n - 2),
```
Mark-Simulacrum, line 20:
```suggestion
        _ => fibonacci(n - 1).await + fibonacci(n - 2).await,
```

Mark-Simulacrum: Or, I guess, maybe I'm confused - these awaits seem necessary, but then this code won't compile as-is.

src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
```diff
@@ -0,0 +1,119 @@
...
+ --> src/lib.rs:1:9
+  |
+1 | #![deny(automatic_box)]
+  |         ^^^^^^^^^^^^^
+help: to request a box explicitly, write `box async fn`
+  |
+4 |     box async fn fibonacci(x: usize) -> usize {
+```
+
+"What's this?" she thinks. "It worked before?" She clicks in her IDE on the "more info" page for the `automatic_box` lint, where she learns that recursive asynchronous functions require some form of extra heap allocation to work. The compiler normally inserts these automatically, but there is an allow-by-default lint that can be used to turn those into warnings or hard errors.
```
rylev, line 48: So, if the lint hadn't been there, the heap allocation would have been added even if Grace hadn't added the explicit `box` annotation? It seems like we'd more likely want to error automatically but add an error message explaining exactly what to do to get the code compiling.

src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
```diff
@@ -0,0 +1,121 @@
...
+  |              ^^^^^^^^^ the function `fibonacci` is recursive and requires a `Box`
+  |
+note: the lint level is defined here
+ --> src/lib.rs:1:9
+  |
+1 | #![deny(automatic_box)]
+  |         ^^^^^^^^^^^^^
+help: to request a box explicitly, write `box async fn`
+  |
+4 |     box async fn fibonacci(x: usize) -> usize {
```
pickfire, line 45: Doesn't this requires changes in the language? Are we going to make an RFC for this later?

src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
nikomatsakis: I disagree, but feel free to write your own story :)

nikomatsakis: I wanted to write a version where people don't have to care about memory allocations, but they have the option to do so. I would however be happy to add some parts of the story describing some of the various downsides one might see from this scenario!

src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
nikomatsakis: Yes, this would require changes to the language, and naturally any change would require an RFC -- or perhaps multiple RFCs! -- before it becomes real. This is a brainstorming exercise. =) What *could* we be working towards?

src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
nikomatsakis: I agree, I think I just overlooked the awaits?

src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
nikomatsakis: (I would though like to read the story where they *do* have to care, as well.)

src/vision/shiny_future/grace_writes_a_recursive_fn_with_auto_box.md
pickfire: One thing with this design is I think it could be confusing. Right now we have, pub, async, const, unsafe, fn and with this we get more, and the ordering matters, at least I had missed `pub async` ordering more than once so I sent a pull request to update it. I think using the same annotation using macros for now might be a better case rather than adding more specific ordering.

nikomatsakis: Yeah. I'm not sure if `box async fn` is right-- I could easily be persuaded to (for example) `#[boxed]` or various other versions. The main thing that is key to the this particular variant would be that there is *some* way to easily introduce an explicit box.

nikomatsakis:
Closing in favor of #222

PR #166, opened 2021-04-21
propose alternative FAQs for shiny futures

nikomatsakis:
Here are some streamlined FAQs for shiny future stories.


nikomatsakis:
Added a question suggested by @sandrmiller about the story having unexpected twists

PR #167, opened 2021-04-21
[shiny future] one possible async drop story

nikomatsakis:


jbr:
Does the use of block_on mean that if multiple contained types implemented AsyncDrop that they'd be run serially in drop order instead of concurrently?

tmandry:
> Does the use of block_on mean that if multiple contained types implemented AsyncDrop that they'd be run serially in drop order instead of concurrently?

This is sort of an assumption of the RAII model, that drops run in sequential well-defined order. It might be interesting to propose a shiny future where they run concurrently instead, and explore the implications of that.

src/vision/shiny_future/alan_cancels_the_future.md
```diff
@@ -0,0 +1,201 @@
...
+## 🤔 Frequently Asked Questions
+
+### What status quo stories are you retelling?
+
+*Link to status quo stories if they exist. If not, that's ok, we'll help find them.*
+
+### What are the key attributes of this shiny future?
+
+* There is an `AsyncDrop` trait that you can implement instead of `Drop`
+    * If you do so, `Drop` will execute your async drop with... `std::block_on`? Something.
```
tmandry, line 153: This implies some level of integration or compatibility between `std::block_on` and the async runtime. Specifically, the reactor needs to run for any kind of I/O to work.

I'm not sure where or whether to point out notes like this in the template, though.

tmandry:
> Work in some part of the story that addresses [the destructor state problem](https://boats.gitlab.io/blog/post/poll-drop/) from boat's blog post. How do we manage generic code that doesn't know whether the future produced by `Drop` is send, and how do we walk about that? Similarly trait objects?

I keep bumping into marker trait problems like this, and I'm wondering how we can "box up" various trait bounds to both make them more convenient and to reason about them generically. My mind keeps moving to the shape of a solution offered by trait aliases and, by extension, associated traits.

This is probably flawed in all kinds of ways but to expound a bit..

**Trait aliases for convenience** (plus anonymous / implicit traits)

```rust
// In a module, we do this (similar to making a `Result` alias) to save ourselves some typing.
trait<T> Future<Output = T> = Future<Output = T> + Send + Sync + 'static;
```

**Associated traits for generics**

```rust
trait AsyncDrop {
    trait Future: Future<Output = ()>;
}

impl AsyncDrop for MyDatabase {
    trait Future = Future<Output = ()> + Send;
}

async fn use_db(db: impl AsyncDrop<Future: Send>>) {
  ...
}
```

Okay, at least that's out of my system for now.

src/vision/shiny_future/alan_cancels_the_future.md
nikomatsakis: Yeah. The interaction between sync drop and async drop is pretty tricky. I'd like to see a few stories exploring different angles on it.

nikomatsakis:
@tmandry I don't quite understand why that is an associated *trait* and not an associated *type*

nikomatsakis:
@jbr 

> Does the use of block_on mean that if multiple contained types implemented AsyncDrop that they'd be run serially in drop order instead of concurrently?

If they were dropped in a synchronous context, yes. What would happen in an asynchronous context is an interesting question! One could imagine the "autogenerated code" using `join`, I hadn't thought about that.

YohDeadfall:
I like the idea of scheduling multiple asynchronous drops in parallel, but definitely wouldn't like to see that to happen implicitly for a simple reason - the compiler cannot observe far that the code allows or that happens in unsafe part. At the same time safe parts are covered by lifetimes and, perhaps, it's possible to determine dependencies between them and produce the right order of drops.

No good example comes to the mind right now, but imagine a guard like object which implements `AsyncDrop`, and some resource registered in that guard. The guard on drop performs some actions involving IO on the resource, so the resource should last till it's used by the guard. Plus there's another independent resource which can be safely dropped at any time. Therefore, we have 3 lifetimes:
* `'r1` for resource 1,
*  `'r2 for resource 2,
*  `'g` for guard and `'g` is less that `'r1`.

In such case the compiler can produce code similar to the next one:

```rust
let drop_res2 = async_drop(res2);

async_drop(guard).await;
async_drop(res1).await;
drop_res2.await;
```

As I said before, it's quite dangerous to determine is it possible to drop objects simultaneously or not. Another reason not to have this behavior turned on by default is limitations of the used runtime, it might not handle high concurrency well or it might give priority to other scheduled tasks leaving async drops in wait queue. This usually leads to resource starvation and perf drops.

nikomatsakis:
Closing in favor of #222

PR #169, opened 2021-04-22
barbara gets burned by select

nikomatsakis:
Based on @tomaka's blog post https://tomaka.medium.com/a-look-back-at-asynchronous-rust-d54d63934a1c

### DIFF
--- a/src/vision/status_quo/barbara_gets_burned_by_select.md
+++ b/src/vision/status_quo/barbara_gets_burned_by_select.md
@@ -51,7 +51,7 @@ Barbara has a hard time figuring out the best way to fix this problem.
 
 * Cancellation doesn't always cancel the entire task; particularly with `select!`, it sometimes cancels just a small piece of a given task.
     * This is in tension with Rust's original design, which was meant to tear down an entire thread or task at once, precisely because of the challenge of writing exception-safe code.
-* Writing "cancellation safe" code is very challenging.
+* Cancellation in Async Rust therefore can require fine-grained recovery.
 
 ### **What are the sources for this story?**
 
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/barbara_gets_burned_by_select.md
```diff
@@ -0,0 +1,76 @@
...
+    * This is in tension with Rust's original design, which was meant to tear down an entire thread or task at once, precisely because of the challenge of writing exception-safe code.
+* Writing "cancellation safe" code is very challenging.
+
+### **What are the sources for this story?**
+
+This was based on [tomaka's blog post](https://tomaka.medium.com/a-look-back-at-asynchronous-rust-d54d63934a1c), which also includes a number of possible solutions, all of them quite grungy.
+
+### **Why did you choose Barbara to tell this story?**
+
+tomaka is a veteran Rust user.
```
pickfire, line 62: Should we put "tomoka" here which is not within the 4 characters?

tomaka: It's fine by me to either mention or not mention me, whatever you decide

rylev:
I’m all for more stories but is this different enough from https://github.com/rust-lang/wg-async-foundations/blob/master/src/vision/status_quo/alan_builds_a_cache.md? It seems to cover the same exact issue without it being specifically about `select!`.

nikomatsakis:
So @rylev I pushed some comments about what makes this story different. I do think that ultimately this story and the others about cancellation would be merged into one tighter narrative, but I'd rather wait to do that until we have assembled all the pieces.

nikomatsakis:
>  It seems to cover the same exact issue without it being specifically about `select!`.

As I tried to explain in the text, the whole motivation for writing this story to me was to *specifically about `select!`*, and the fact that cancellation of futures requires fine-grained recovery even for normal, expected events. It's not just about a "the user dropped the connection, so tear down everything related to it", which is how I thought of it to start.

src/vision/status_quo/barbara_gets_burned_by_select.md
```diff
@@ -51,20 +51,28 @@ Barbara has a hard time figuring out the best way to fix this problem.
...
 ### **Why did you choose Barbara to tell this story?**
 
-tomaka is a veteran Rust user.
+The problem described here could strike anyone, including veteran Rust users. It's a subtle interaction that is independent of source language. Also, the original person who reported it, tomaka, is a veteran Rust user.
 
 ### **How would this story have played out differently for the other characters?**
 
 They would likely have a hard time diagnosing the problem. It really depends on how well they have come to understand the semantics of cancellation. This is fairly independent from programming language background; knowing non-async Rust doesn't help in particular, as this concept is specific to async code.
 
+### What is different between this story and other cancellation stories?
```
rylev, line 68: This section is fantastic. It's very subtle, but it has big implications. ♥

rylev (approved):

PR #170, opened 2021-04-22
Add link to shiny future blog post

seanmonstar:
When I went to check on some details of the shiny future purpose, I eventually found that the blog post was very useful, but wasn't linked on the shiny future index page. This is just a suggested addition :shrug:


tmandry:
Thanks!

PR #173, opened 2021-04-26
extend end of brainstorming deadline to 2021-05-14

nikomatsakis:
We've been doing an amazing job with the status quo stories, but I think we're going to want more time to kick around shiny future. Giving us a bit more time.


seanmonstar (approved):

zeenix (approved):
Sounds great to me.

PR #174, opened 2021-04-26
shiny future: Barbara enjoys an async sandwich

seanmonstar:
[Rendered](https://github.com/seanmonstar/wg-async-foundations/blob/shiny-future-barbara-async-sandwich/src/vision/shiny_future/barbara_enjoys_an_async_sandwich.md)

### DIFF
--- a/src/vision/shiny_future/barbara_enjoys_an_async_sandwich.md
+++ b/src/vision/shiny_future/barbara_enjoys_an_async_sandwich.md
@@ -1,0 +1,160 @@
+# ✨ Shiny future stories: Barbara enjoys her async-sync-async sandwich :sandwich: 
+
+:::warning
+Alternative titles:
+- Barbara enjoys her async-sync-async sandwich :sandwich: 
+- Barbara recursively blocks
+- Barbara blocks and blocks and blocks
+:::
+
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "shiny future" story submitted as part of the brainstorming period. It is derived from what actual Rust users wish async Rust should be, and is meant to deal with some of the challenges that Async Rust programmers face today.
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as peoples needs and desires for async Rust may differ greatly, shiny future stories [cannot be wrong]. At worst they are only useful for a small set of people or their problems might be better solved with alternative solutions). Alternatively, you may wish to [add your own shiny vision story][htvsq]!
+
+## The story
+
+Barbara wants to customize a permissions lookup when accepting requests. The library defines a `trait PermitRequest`, to allow the user to define their own rules. Nice!
+
+```rust
+trait PermitRequest {}
+```
+
+She starts small, to get her feet wet.
+
+```rust
+struct Always;
+
+impl PermitRequest for Always {
+    fn permit(&self, _: &Request) -> bool {
+        true
+    }
+}
+```
+
+All requests are permitted! Simple, but now to actually to implement the permissions logic.
+
+One of the basic rules Barbara has is to check the request for the existence of a header, but the function is written as `async`, since Barbara figured it might need to be eventually.
+
+```rust
+async fn req_has_header(req: &Request) -> bool {
+    req.headers().contains_key("open-sesame")
+}
+```
+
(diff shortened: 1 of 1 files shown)


nikomatsakis (approved):
I like this, esp. the hints about some of the ways it might be done. Thanks @seanmonstar!

pickfire:
Just wondering, now the difference is that `block_on` is able to run within `sync` code.

What if the user did not create any runtime but `block_on` was still called? Will the code get a runtime panic instead of compile time error?

Maybe we could have a global `OnceCell` and `GhostCell` (feels like global variable) which can be enforced at compile time, to determine whether a runtime is being created? Maybe it could solve the runtime not found (runtime not created), runtime within runtime (runtime on top of runtime) and runtime feature not enabled (like io feature not enabled) issues with errors at compile time?

nikomatsakis:
> What if the user did not create any runtime but block_on was still called? Will the code get a runtime panic instead of compile time error?

This is a good question to highlight. What happens in this scenario?

Perhaps also in a case where there are multiple runtimes.

seanmonstar:
> What if the user did not create any runtime but block_on was still called? Will the code get a runtime panic instead of compile time error?

I imagined that calling `block_on` would have been basically what the `#[runtime::main]` attributes do, so it would also Just Work. It would however be more expensive then if someone set it up once. It might be possible to have it check a thread-local to prevent multiple creations, I haven't decided if this is a good or bad idea though.

> Perhaps also in a case where there are multiple runtimes.

Could be an interesting separate story. There would be a lot of nuance to getting the implementation and decisions of that right. The original idea was that runtimes would all use `futures::executor::enter()` so that if mixed, they could at least detect the mixture and yell that it's a bad idea.

pickfire:
> I imagined that calling block_on would have been basically what the #[runtime::main] attributes do, so it would also Just Work. It would however be more expensive then if someone set it up once. It might be possible to have it check a thread-local to prevent multiple creations, I haven't decided if this is a good or bad idea though.

Something like glibc runtime feature detection which is only done once and cached? See https://github.com/rust-lang/rust/issues/68455#issuecomment-824114443

src/vision/shiny_future/barbara_enjoys_an_async_sandwich.md
```diff
@@ -0,0 +1,160 @@
+# ✨ Shiny future stories: Barbara enjoys her async-sync-async sandwich :sandwich: 
+
+:::warning
+Alternative titles:
```
seanmonstar, line 4: I wasn't certain about the title, so I listed some alternatives. Now a few days later, I still prefer the original, but I can change if desired. Otherwise, I can just nuke this part first.

src/vision/shiny_future/barbara_enjoys_an_async_sandwich.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+When Barbara goes to implement the `PermitRequest` trait, she realizes a problem: the trait did not think permissions would require an async lookup, so its method is not `async`. Barbara tries the easiest thing first, hoping that she can just block on the future.
+
+```rust
+struct HasHeader;
+
+impl PermitRequest for HasHeader {
+    fn permit(&self, req: &Request) -> bool {
+        task::block_on(req_has_header(req))
```
zeenix, line 54: While a `block_on` in std would already go a long way here, given how often this is needed, I wonder if we could do even better and provide a method on futures, so people can just use it like `.await`:

```rust
        req_has_header(req).block()
```

In fact, if it was up to me, I'd even want it as part of the language:

```rust
    req_has_header(req).block
```

seanmonstar: The original `Future` trait had a `wait` method, which did exactly that. We chose to remove it, because it was _too easy_ for people to reach for instead of just letting the future be chained with others (now I'd say just `.await`). While I want to make it more forgiving in instances when a user _has_ to do this, I'd still want to encourage users to try not to, since it _is_ worse than using `.await`.

zeenix: hmm.. `wait` can very easily be confused with `await` and I can see most programmers new to async, confusing them, but I see your point.

seanmonstar: Right, I'm referring to `Future` v0.1, which existed before there was async/await syntax. People would often do things like this because of how easy it was:

```rust
first_fut.and_then(|val| {
    // a few lines of whatever
    let foo = do_another_future(val).wait();
    // and other lines
})
```

PR #176, opened 2021-04-29
status quo story: Barbara wishes for an easy runtime switch

zeenix:
This story is the outcome of our status-quo brainstorming session on Wed Apr 28, 2021.

### DIFF
--- a/src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
+++ b/src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
@@ -25,7 +25,7 @@ base:
 
 * tokio provides variants of helpers macros and types, like `tokio::select!` and `tokio::Mutex`.
   These helpers can be used without the rest of tokio, and there are also alternatives from the
-  `futures crate and elsewhere (albeit with subtle differences).
+  `futures` crate and elsewhere (albeit with subtle differences).
 * tokio uses a custom version of `AsyncRead` and `AsyncWrite` traits which differ from the ones used
   by other parts of the ecosystem.
 * The tokio API is needed to create core runtime operations like timers (`tokio::time::sleep`) and
--- a/src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
+++ b/src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
@@ -13,12 +13,9 @@ quite large (in the millions of lines of code). They've been relying on tokio as
 and the codebase makes heavy use of its rich API. It has served them well over the years and they're
 very happy with it.
 
-Meanwhile, another team in Barbara's company have developed another large project, where they chosen
-async-std as their runtime of choice. This team presents their project in a company meeting and it
-gets Barbara excited. She starts wondering about possible collaboration and code-reuse between the
-two projects. She soon realizes that such collaboration would be a lot easier if both projects used
-the same runtime. She decides to look into the possibility of porting her projects from tokio to
-async-std.
+Barbara knows about async-std but has never used it. She has wondered for a while how her
+application would work and perform if she had used async-std instead. She decides to test it out by
+porting her projects from tokio to async-std.
 
 To their disappointment, they discover many areas, where their choice of runtime permeates the code
 base:
--- a/src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
+++ b/src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
@@ -1,0 +1,76 @@
+# 😱 Status quo stories: Barbara wishes for an easy runtime switch
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from
+real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async
+Rust programmers face today.
+
+## The story
+
+[Barbara] has been working on an async codebase for the past 5 years. It is extremely mature and
+quite large (in the millions of lines of code). They've been relying on tokio as their async runtime
+and the codebase makes heavy use of its rich API. It has served them well over the years and they're
+very happy with it.
+
+Meanwhile, another team in Barbara's company have developed another large project, where they chosen
+async-std as their runtime of choice. This team presents their project in a company meeting and it
+gets Barbara excited. She starts wondering about possible collaboration and code-reuse between the
+two projects. She soon realizes that such collaboration would be a lot easier if both projects used
+the same runtime. She decides to look into the possibility of porting her projects from tokio to
+async-std.
+
+To their disappointment, they discover many areas, where their choice of runtime permeates the code
+base:
+
+* tokio provides variants of helpers macros and types, like `tokio::select!` and `tokio::Mutex`.
+  These helpers can be used without the rest of tokio, and there are also alternatives from the
+  `futures crate and elsewhere (albeit with subtle differences).
+* tokio uses a custom version of `AsyncRead` and `AsyncWrite` traits which differ from the ones used
+  by other parts of the ecosystem.
+* The tokio API is needed to create core runtime operations like timers (`tokio::time::sleep`) and
+  to launch tasks; there doesn't seem to be a standard way to abstract over those kinds of things in
+  a runtime-independent way.
+* Some of their dependencies (e.g `hyper` and `reqwest`) are tied to tokio. In some cases, there are
+  configuration options or ways to use those dependencies that don't depend on tokio, but there is
+  no standard mechanism for that.
+
+These things aren't specific to tokio. There just doesn't seem to be a lot of consensus in the
+ecosystem on how to write "runtime-independent" code and in some cases there aren't any great
+options available (e.g., spawning tasks).
+
+They investigate the possibility of providing some sort of compatibility layer between tokio
+and their new runtime of choice but this turns out to not seem like the right way to go as this
+compatibility layer would require too much overhead.
+
+Realizing that the task of porting the entire code base to async-std, will take a lot of effort and
+time, Barbara decides to give up. She is very disappointed.
+
+## 🤔 Frequently Asked Questions
+
+### **What are the morals of the story?**
+* Using a certain executor often means using a certain run-time ecosystem. This often locks the user
+  into that ecosystem.
+* Tying yourself to a certain executor means that you are tied to the priorities of that executor.
+  You may be happy with the run-time ecosystem, but have special needs that the default executor
+  does not provide. If the executor doesn't have an extensibility model, you're stuck. **Note:**
+  It is perfectly reasonable for a general purpose executor to not be able or willing to cater for
+  specialized needs.
+* All of this is made worse by that fact that [run-time agnostic libraries] are difficult and
+  sometimes even impossible to write.
+
+### **What are the sources for this story?**
+
+This story is more of a thought experiment than a recounting of a true story. We just asked
+logically what would happen if a team working on code base where it was assumed they could use a
+specific runtime decides to use a different runtime.
+
+### **Why did you choose Barbara to tell this story?**
+The story assumes a Rust programmer that has worked for several years on a large and complex Rust
+codebase, so Barbara is the natural choice here.
+
+### **How would this story have played out differently for the other characters?**
+It wouldn't. If this story happens them, they're on the same level of Rust expertise as Barbara is.
+
+[Barbara]: ../characters/barbara.md
+[run-time agnostic libraries]: https://github.com/rust-lang/wg-async-foundations/issues/45


zeenix:
cc @rylev.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,59 @@
...
+the low-end edge devices Barbara is targetting, memory consumption is even a hard constraint there.
+Additionally, the usage of a threadpool by runtime X does not help either as the targetted edge
+devices have either a single CPU thread or only a few.
+
+They had never planned to run on edge devices and so it was assumed that they would always stick
+to runtime X. Barbara looks into the how much work it would be to convert their codebase from
+runtime X to a custom runtime tuned to their workload. To their disapointment, they discover many
+areas, where their choice of runtime permeates the code base:
+
+* They extensively use runtime-dependent helpers (e.g., a select! macro).
```
rylev, line 27:
```suggestion
* They extensively use runtime-dependent helpers (e.g., a `select!` macro).
```

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,59 @@
...
+
+### **What are the morals of the story?**
+* Using a certain executor often means using a certain run-time ecosystem. This often locks the user into that ecosystem.
+* Tying yourself to a certain executor means that you are tied to the priorities of that executor.
+    * You may be happy with the run-time ecosystem, but have special needs that the default executor does not provide. If the executor doesn't have an extensibility model, you're stuck.
+* All of this is made worse by that fact that [run-time agnostic libraries](https://github.com/rust-lang/wg-async-foundations/issues/45) are difficult and sometimes impossible to write.
+
+### **What are the sources for this story?**
+
+**TODO**
```
rylev, line 51: This story is certainly more of a thought experiment than a recounting of some true story. I think it's fine to say we just asked logically what would happen if a team working on code base where it was assumed they could use a specific runtime decides to use a different runtime.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,59 @@
...
+extremely expensive and takes much longer than she originally thought. Barbara is very disapointed.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+### **What are the morals of the story?**
+* Using a certain executor often means using a certain run-time ecosystem. This often locks the user into that ecosystem.
+* Tying yourself to a certain executor means that you are tied to the priorities of that executor.
+    * You may be happy with the run-time ecosystem, but have special needs that the default executor does not provide. If the executor doesn't have an extensibility model, you're stuck.
```
rylev, line 46: I would add the point that it's reasonable that general purpose executors won't always be willing or able to handle specialized needs.

rylev (commented):
♥ Thanks!

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
nikomatsakis: I agree with that.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,59 @@
...
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from
+real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async
+Rust programmers face today.
+
+## The story
+
+[Barbara] has been working on an async codebase for the past 5 years. It is extremely mature and
+quite large (in the millions of lines of code). The product has traditionally been run on the server
+only, but now they want to move large parts to the edge. After some investigation, they realize that
+the open-source async runtime they have been using (runtime X), while great for server workloads, is
```
nikomatsakis, line 14: This says "runtime X", but I think it's really about `tokio`...? We've generally tried to be real and concrete, rather than talking in hypotheticals. I think we should stick to that.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,59 @@
...
+They investigate the possibility of providing some sort of compatibility layer between runtime X
+and their new runtime of choice but this turns out to not seem like the right way to go as this
+compatibility layer would require too much overhead.
+
+Barbara makes the decision to rewrite large parts of the codebase to use the new runtime. This is
+extremely expensive and takes much longer than she originally thought. Barbara is very disapointed.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
```
nikomatsakis, line 41: Can you remove this line?

nikomatsakis (changes requested):
A few thoughts, but this seems pretty good to me.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
rylev: This was my suggestion since I didn't want to make it seem like this is an issue with how tokio is designed, but as long as we make a mention in the FAQ that we see the same thing happening (to different degrees) with all runtimes, I'm happy with the idea of switching to use tokio in this example.

nikomatsakis: I would definitely include a comment in the FAQ

nikomatsakis: To me, referring "obliquely" to a runtime makes it read sinister or something.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,59 @@
...
+Additionally, the usage of a threadpool by runtime X does not help either as the targetted edge
+devices have either a single CPU thread or only a few.
+
+They had never planned to run on edge devices and so it was assumed that they would always stick
+to runtime X. Barbara looks into the how much work it would be to convert their codebase from
+runtime X to a custom runtime tuned to their workload. To their disapointment, they discover many
+areas, where their choice of runtime permeates the code base:
+
+* They extensively use runtime-dependent helpers (e.g., a select! macro).
+* The timers, tasks, and async locks implementation all assume runtime X.
```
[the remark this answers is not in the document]
zeenix: I had a similar understanding/feeling but I thought maybe I'm missing something so didn't object much to these points. @nikomatsakis @rylev ?

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
zeenix: OK, so I'll change this to `tokio`.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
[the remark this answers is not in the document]
nikomatsakis: Well, as a simple example, `tokio::select` is a specific `select` macro, and tokio has async locks that are closely integrated with its runtime...? I believe other runtimes offer similar facilities. I think that's what I assumed was being referred to here.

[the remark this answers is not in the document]
zeenix: > tokio has async locks that are closely integrated with its runtime

Yeah, I thought so too but recently @taiki-e informed me that that's not correct and you can build tokio with only the `sync` module w/o the rest.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
zeenix: @rylev thanks. I'll actually add what you wrote here. :)

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,66 @@
...
+devices have either a single CPU thread or only a few.
+
+They had never planned to run on edge devices and so it was assumed that they would always stick
+to tokio. Barbara looks into the how much work it would be to convert their codebase from
+tokio to a custom runtime tuned to their workload. To their disapointment, they discover many
+areas, where their choice of runtime permeates the code base:
+
+* They extensively use runtime-dependent helpers (e.g., a `select!` macro).
+* The timers, tasks, and async locks API is all tokio-specific.
+* Above all else, tokio uses a custom version of `AsyncRead` and `AsyncWrite` traits which
```
nikomatsakis, line 29: I think another factor would be that many of their dependencies are tied to tokio, which probably links to other stories.

nikomatsakis (changes requested):

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
zeenix: You mean `AsyncRead` and `AsyncWrite`? The timers and async locks [aren't at least](https://github.com/rust-lang/wg-async-foundations/pull/176#discussion_r625670427).

nikomatsakis: I meant e.g. `reqwest` or `hyper`

zeenix: ah right, yeah that'd make sense.

zeenix:
All threads resolved. I'll squash the fixups now..

zeenix:
> All threads resolved. I'll squash the fixups now..

Done.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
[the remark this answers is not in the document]
taiki-e: > recently @taiki-e informed me that that's not correct and you can build tokio with only the `sync` module w/o the rest.

Yeah, they don't depend on a specific runtime.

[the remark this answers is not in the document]
taiki-e: (Btw, @zeenix marked this conversation as resolved, but I reverted it to unresolved because the relevant text hadn't been updated and seems not correct.)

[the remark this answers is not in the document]
zeenix: @taiki-e thanks. I actually did but somehow that change got lost locally. :( I'll update again..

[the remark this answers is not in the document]
zeenix: @taiki-e actually, I did fix it: it says:

> The timers, tasks, and async locks API is all tokio-specific,

While they're not tied to the runtime itself, they **are** tokio-specific in the sense that it's all custom API and not implementing some traits from stardardized/common traits.

I also added:

>  even if they're not tied to the tokio runtime and can be used independently of it.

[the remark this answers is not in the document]
taiki-e: > The timers, tasks, and async locks API is all tokio-specific,

Timers and tasks actually depend on runtime, so I think this is misleading.

> While they're not tied to the runtime itself, they **are** tokio-specific in the sense that it's all custom API and not implementing some traits from stardardized/common traits.

I'm not sure what you are talking about. What are the standardized/common traits used in async lock primitives? (I don't know of any other traits that are used for that purpose other than `Future`.)

Also, what do you mean by "custom API"?

> I did fix it

I don't think `select` related thing was fixed.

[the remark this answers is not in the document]
rylev: Well they are runtime dependent in a way - you cannot use the `tokio::select` macro in a project that doesn't use `tokio` without bringing `tokio` in as a dependency. So it's not quite as bad as being directly dependent on the runtime in order to function, but there is still a dependency relationship.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+## The story
+
+[Barbara] has been working on an async codebase for the past 5 years. It is extremely mature and
+quite large (in the millions of lines of code). The product has traditionally been run on the server
+only, but now they want to move large parts to the edge. After some investigation, they realize that
+the open-source async runtime they have been using (tokio), while great for server workloads, is
+not very performant on the low-end edge devices.
+
+Tokio makes tradeoffs in memory for performance and while performance is still a concern on
+the low-end edge devices Barbara is targetting, memory consumption is even a hard constraint there.
```
rylev, line 18:
```suggestion
the low-end edge devices Barbara is targeting, memory consumption is an even harder constraint there.
```

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+Additionally, the usage of a threadpool by tokio does not help either as the targetted edge
```
rylev, line 19:
```suggestion
Additionally, the usage of a thread pool by tokio does not help either as the targeted edge
```

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+devices have either a single CPU thread or only a few.
+
+They had never planned to run on edge devices and so it was assumed that they would always stick
+to tokio. Barbara looks into the how much work it would be to convert their codebase from
+tokio to a custom runtime tuned to their workload. To their disapointment, they discover many
```
rylev, line 24:
```suggestion
tokio to a custom runtime tuned to their workload. To their disappointment, they discover many
```

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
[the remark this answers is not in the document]
taiki-e: @rylev It would be quite misleading to rephrase "depends on the library providing the runtime" to "depends on the runtime".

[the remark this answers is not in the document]
zeenix: > > The timers, tasks, and async locks API is all tokio-specific,
> 
> Timers and tasks actually depend on runtime, so I think this is misleading.

Right, need to break this line then.



> > While they're not tied to the runtime itself, they **are** tokio-specific in the sense that it's all custom API and not implementing some traits from stardardized/common traits.
> 
> I'm not sure what you are talking about. What are the standardized/common traits used in async lock primitives? (I don't know of any other traits that are used for that purpose other than `Future`.)

There are none, that's the problem, not that tokio doesn't implement them. Please keep in mind that the point is not to criticize tokio here (and hence the reason we were reluctant to even mention it here).

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
zeenix: always get `targeting` spelling wrong (even just now, had to correct it). :smile:

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+* They extensively use runtime-dependent helpers (e.g., a `select!` macro).
+* The timers, tasks, and async locks API is all tokio-specific, even if they're not tied to the
+  tokio runtime and can be used independently of it.
+* Some of their dependencies (e.g `hyper` and `reqwest`) are tied to tokio.
+* Above all else, tokio uses a custom version of `AsyncRead` and `AsyncWrite` traits which
+  differ from the ones used by rest of the ecosystem.
+
+They investigate the possibility of providing some sort of compatibility layer between tokio
+and their new runtime of choice but this turns out to not seem like the right way to go as this
+compatibility layer would require too much overhead.
```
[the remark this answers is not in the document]
nikomatsakis: This is an attempt to incorporate the clarifications from @taiki-e and structure these bullets a bit more. I'm curious how it reads for folks. =)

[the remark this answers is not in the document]
nikomatsakis: I also think we could link to other stories for each bullet? I feel like we have some on each theme.

nikomatsakis (changes requested):
Thoughts?

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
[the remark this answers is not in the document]
nikomatsakis: I took a stab at writing something below. I do think that needing a dependency on tokio is worth nothing; I also did not know that tokio-util can be used independently, and it seems like others didn't (I didn't see a lot of docs). So i think it's valid for Barbara in the story to be confused about this, but we should totally clarify the reality of the situation as well.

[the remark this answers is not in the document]
nikomatsakis: I do wonder how the nifty [preemption stuff](https://tokio.rs/blog/2020-04-preemption) works with mutexes when you're not using the tokio runtime for other things! I'd like to read more into the tokio codebase...

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+[Barbara] has been working on an async codebase for the past 5 years. It is extremely mature and
+quite large (in the millions of lines of code). The product has traditionally been run on the server
+only, but now they want to move large parts to the edge. After some investigation, they realize that
+the open-source async runtime they have been using (tokio), while great for server workloads, is
+not very performant on the low-end edge devices.
+
+Tokio makes tradeoffs in memory for performance and while performance is still a concern on
+the low-end edge devices Barbara is targeting, memory consumption is an even harder constraint there.
+Additionally, the usage of a thread pool by tokio does not help either as the targeted edge
+devices have either a single CPU thread or only a few.
```
Darksonn, line 20: I think that it is worth noting that Tokio has a different runtime configuration for the single-threaded use case.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+
+## The story
+
+[Barbara] has been working on an async codebase for the past 5 years. It is extremely mature and
+quite large (in the millions of lines of code). The product has traditionally been run on the server
+only, but now they want to move large parts to the edge. After some investigation, they realize that
+the open-source async runtime they have been using (tokio), while great for server workloads, is
+not very performant on the low-end edge devices.
+
+Tokio makes tradeoffs in memory for performance and while performance is still a concern on
```
carllerche, line 17: Tokio has low-resource modes, it would be helpful to be more precise around what specifically Tokio does that doesn't work.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+* Some of their dependencies (e.g `hyper` and `reqwest`) are tied to tokio.
+* Above all else, tokio uses a custom version of `AsyncRead` and `AsyncWrite` traits which
+  differ from the ones used by rest of the ecosystem.
+
+They investigate the possibility of providing some sort of compatibility layer between tokio
+and their new runtime of choice but this turns out to not seem like the right way to go as this
+compatibility layer would require too much overhead.
+
+Barbara makes the decision to rewrite large parts of the codebase to use the new runtime. This is
+expensive and takes longer than she originally thought. Barbara is very disapointed.
```
jbr, line 39:
```suggestion
expensive and takes longer than she originally thought. Barbara is very disappointed.
```

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+
+## 🤔 Frequently Asked Questions
+
+### **What are the morals of the story?**
+* Using a certain executor often means using a certain run-time ecosystem. This often locks the user
+  into that ecosystem.
+* Tying yourself to a certain executor means that you are tied to the priorities of that executor.
+  You may be happy with the run-time ecosystem, but have special needs that the default executor
+  does not provide. If the executor doesn't have an extensibility model, you're stuck. **Note:**
+  It is perfectly reasonable for a general purpose executors to not be able or willing to cater for
```
jbr, line 49:
```suggestion
  It is perfectly reasonable for a general purpose executor to not be able or willing to cater for
```

jbr (commented):
Small copyedits while reading

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
jbr: It seems like this is a contrived example specifically intended to describe some arbitrary capability tokio doesn't support, along the lines of the second bullet in the morals of the story section. Maybe TTCP from [alan switches runtimes](https://github.com/rust-lang/wg-async-foundations/blob/master/src/vision/shiny_future/alan_switches_runtimes.md) would provide a clearer example?

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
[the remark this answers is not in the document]
carllerche: @rylev you *can* bring in the macros, io traits, etc... w/o the runtime by depending on tokio without specifying the runtime flag. Tokio avoids default features specifically for this case.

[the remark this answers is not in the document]
carllerche: When not using Tokio types w/ the tokio runtime, there is no preemption and the types behave as one would otherwise expect.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
carllerche: If I understand your point, you are suggesting that the reason for wanting to switch is hypothetical and not necessarily representative of an actual story? I also looked at the TTCP story. In that case, I am able to translate TTCP to a real use case (e.g. QUIC). In this case, I am not. The story describes starting with Tokio, but then hitting some hypothetical problem.

The reason I bring this up is because **if** there is a real blocker w/ Tokio, then the "shiny future" story could be to fix the problem and then there is no reason to switch :) but the way this status quo story is written doesn't leave that option open.

jbr: Exactly. Although I didn't write this story, I have similar concerns with the status quo that start from the assumption outlined in the second bullet point of the moral of the story section: That one of the _benefits_ of async rust's runtime being external to std is that there can and will be different user needs which don't need to be met by one single executor. If the tokio design goal is to support all possible use cases of a [std] async executor as well as a purpose built executor could, that would a) be super exciting and b) would universally obviate the "switching runtimes" concern by treating any need for a different executor as an issue for tokio to introduce the right extension point

carllerche: I can say that Tokio would like to support as many use cases as it can :) I believe that something like [Embassy](https://github.com/embassy-rs/embassy) will always be out of scope, but I also believe a user of Embassy would have to build their app assuming Embassy from the start.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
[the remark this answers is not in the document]
zeenix: @nikomatsakis LGTM but I believe there is an extra `and` here?

> tokio provides variants of helpers macros and types, like `tokio::select!` and like `tokio::Mutex`.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
zeenix: > Tokio has low-resource modes, it would be helpful to be more precise around what specifically Tokio does that doesn't work.

We didn't know of that. :) Could you please point to the relevant docs? As @jbr pointed out (I think?), while the point of this story was to cover the problems of switching a runtime, we may not have been particularly successful at coming up with compelling reasons for Barbara to want to switch the runtime.

@nikomatsakis Perhaps we should switch back to our hypothetical runtime?

Darksonn: The main piece of relevant documentation is the [`tokio::runtime::Builder::new_current_thread`][1] function, which lets you create a runtime that doesn't spawn any threads. Additionally, the `#[tokio::main]` macro also has a configuration option for using the current thread runtime, which you can find on the documentation for the macro.

[1]: https://docs.rs/tokio/1/tokio/runtime/struct.Builder.html#method.new_current_thread

zeenix: Since threads were addressed already elsewhere, I understood @carllerche meant something more than just threads with `Tokio has low-resource modes`. Hence my question.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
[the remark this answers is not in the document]
nikomatsakis: yeah. I tweaked it a bit more and made it clearer that indeed some of those utilities are usable without the tokio runtime.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
nikomatsakis: I don't think switching back to a hypothetical runtime really helps that much. The question is why people want to switch. I will say that I've talked to quite a few people who express a desire to *have the option* to switch and experiment with other runtimes. Perhaps the story should be tweaked to be Barbara hearing about a new runtime -- or wanting to implement her own to see if she could do better -- and being curious to see what it would entail, then realizing it would be tricky and not doing it. That seems to be more common in practice. =)

zeenix: @nikomatsakis hmm.. good point. 

> Perhaps the story should be tweaked to be Barbara hearing about a new runtime -- or wanting to implement her own to see if she could do better

Sure but neither of those sound like compelling enough reasons to put a lot of effort and time into this. Barbara will likely just give-up as soon as they realize it's not a trivial endeavor. Although the story doesn't have to be super realistic. :)

How about they needed to add a new feature, which would be a lot easier to add using some existing libraries that are closely tied to another runtime?

jbr: What about Barbara's company making a native desktop version of her app and realizing that there are performance gains from using an os-specific runtime like grand central dispatch?

nikomatsakis: Personally I think it's sufficient to just say that they were investigating what it would take to switch. Maybe she was asked to assess it for her job so she did a good job digging into all the details.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+They had never planned to run on edge devices and so it was assumed that they would always stick
+to tokio. Barbara looks into the how much work it would be to convert their codebase from
+tokio to a custom runtime tuned to their workload. To their disappointment, they discover many
+areas, where their choice of runtime permeates the code base:
+
+* tokio provides variants of helpers macros and types, like `tokio::select!` and `tokio::Mutex`. These helpers can be used without the rest of tokio, and there are also alternatives from the futures crate and elsewhere (albeit with subtle differences).
+    * tokio uses a custom version of `AsyncRead` and `AsyncWrite` traits which differ from the ones used by other parts of the ecosystem.
+* The tokio API is needed to create core runtime operations like timers (`tokio::time::sleep`) and to launch tasks; there doesn't seem to be a standard way to abstract over those kinds of things in a runtime-independent way.
+* Some of their dependencies (e.g `hyper` and `reqwest`) are tied to tokio. In some cases, there are configuration options or ways to use those dependencies that don't depend on tokio, but there is no standard mechanism for that.
```
emmanuelantony2000, line 30: Just wanted to add my two cents here. Afaik `reqwest`, although built on top `hyper` doesn't expose the `Executor` trait. But `hyper` is runtime agnostic, and you could write your own connector for it, and make a client with a custom executor.

emmanuelantony2000 (approved):

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,68 @@
...
+
+These things aren't specific to tokio. There just doesn't seem to be a lot of consensus in the ecosystem on how to write "runtime-independent" code and in some cases there aren't any great options available (e.g., spawning tasks).
+
+They investigate the possibility of providing some sort of compatibility layer between tokio
+and their new runtime of choice but this turns out to not seem like the right way to go as this
+compatibility layer would require too much overhead.
+
+Barbara makes the decision to rewrite large parts of the codebase to use the new runtime. This is
+expensive and takes longer than she originally thought. Barbara is very disappointed.
```
eminence, line 39: Maybe this story doesn't need to end with Barbara rewriting stuff to work with another runtime.  I think the story can still stand if it concludes with Barbara being unsure if making the switch away from tokio is worth it, and being disappointed at the amount of work needed to make the switch.  This would be a bit more consistent with @nikomatsakis's proposal above ("Personally I think it's sufficient to just say that they were investigating what it would take to switch. Maybe she was asked to assess it for her job so she did a good job digging into all the details.")

That said, my comment here is a pretty minor point.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
zeenix: @emmanuelantony2000 Thanks, they're just random examples though. Are there better ones that **are** tied to runtimes?

zeenix: > Just wanted to add my two cents here. Afaik `reqwest`, although built on top `hyper` doesn't expose the `Executor` trait.

Not sure that matters here. If they've a hard dep on `hyper` and `hyper` has a hard dep on tokio, this is not ideal at all.

> But `hyper` is runtime agnostic, and you could write your own connector for it, and make a client with a custom executor.

From what I understand from the docs, it provides an agnostic API but it itself still relies on tokio unconditionally. If the task at hand is to switch to another runtime, this is a problem.

emmanuelantony2000: > @emmanuelantony2000 Thanks, they're just random examples though. Are there better ones that **are** tied to runtimes?

I think `reqwest` is a good example as it is strongly tied to tokio. I don't think it exposes an executor API, like hyper.

> Not sure that matters here. If they've a hard dep on `hyper` and `hyper` has a hard dep on tokio, this is not ideal at all.
> 
> From what I understand from the docs, it provides an agnostic API but it itself still relies on tokio unconditionally. If the task at hand is to switch to another runtime, this is a problem.

It is not an easy switch. But if I'm correct, hyper just depends on the AsyncWrite and AsyncRead traits in tokio, not the runtime itself.

Probably it could be something like "hyper although runtime agnostic, required you to write a Connector if you want to use a custom runtime."

zeenix: > Probably it could be something like "hyper although runtime agnostic, required you to write a Connector if you want to use a custom runtime."

The problem is that even if it'd make it super easy to use other runtimes, it still depends on tokio itself and if the goal is to switch to another runtime, it's not nice to still depend on tokio indirectly and sort of beats the point of the exercise.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
nikomatsakis: I agree, I thnk the story is stronger that way, it represents the real story i've heard most of the time.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
rylev: As a bit of context - this also comes from teams I've talked to who have expressed the desire to not be locked into a specific runtime. We have many happy users of tokio and many happy users of custom runtimes, and there are many teams who are unsure where tokio will always be able to meet their needs or whether they will eventually want to switch off. 

In short, this story wasn't motivated by a real world situation that exists yet (after all, async Rust is fairly new and so we're just not at a point where this might appear). I think it's reasonable to want to ensure that switching off of a certain runtime is doable.

zeenix: I came up with an alternative motivation (something more likely IMO). Please have a look. If people still disagree, it can be changed again of course. :)

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,76 @@
...
+Rust programmers face today.
+
+## The story
+
+[Barbara] has been working on an async codebase for the past 5 years. It is extremely mature and
+quite large (in the millions of lines of code). They've been relying on tokio as their async runtime
+and the codebase makes heavy use of its rich API. It has served them well over the years and they're
+very happy with it.
+
+Meanwhile, another team in Barbara's company have developed another large project, where they chosen
```
rylev, line 16: I'm worried this will shift the story too much towards runtime agnostic libraries. If it's easy to write runtime agnostic libraries than this is a non issue. However, the original intent of the story was to motivate the need for runtime agnostic application code. These are related but IMO distinct issues.

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
```diff
@@ -0,0 +1,76 @@
...
+two projects. She soon realizes that such collaboration would be a lot easier if both projects used
+the same runtime. She decides to look into the possibility of porting her projects from tokio to
+async-std.
+
+To their disappointment, they discover many areas, where their choice of runtime permeates the code
+base:
+
+* tokio provides variants of helpers macros and types, like `tokio::select!` and `tokio::Mutex`.
+  These helpers can be used without the rest of tokio, and there are also alternatives from the
+  `futures crate and elsewhere (albeit with subtle differences).
```
rylev, line 28: Missing "`" at the end of futures

src/vision/status_quo/barbara_wishes_for_easy_runtime_switch.md
zeenix: Good point but even if it was possible and easy to write runtime-agnostic libraries, does it mean everyone would?

nikomatsakis: It seems like it's good enough for her to just wonder "hmm, maybe I should try out async-std for my application"...

zeenix: PTALA.

rylev (commented):
Thanks for riffing on this. I think what we have now is good. The conversation steered a bit too close to a technical discussion of what tokio can and cannot do which wasn't really the point of the story. There will be use cases where people want to switch executors, and I think it's fine to point out that executors often come packaged with runtimes and associated helpers which make switching harder. A potential shiny future of this will simply be that most functionality exists in the std lib and many of the application's dependencies are runtime independent so that it's easier to make a switch if one wants to.

carllerche:
I think, it is probably worth noting, that a similar story happens pretty often with blocking/threads vs. async, and in both directions (wrote sync, would like to switch easily to async and wrote async, but would like to switch to sync).

nikomatsakis:
I'm pretty happy with this. I agree with @carllerche that there all kinds of other cases where people have to really bottom out a complex design to decide if things work better -- sync vs async being the most fundamental -- but this story can't cover them all. =)

PR #188, opened 2021-05-01
Update acknowledgements

nikomatsakis:
I am pretty sure there's a bug in the script though! For example, @Mark-Simulacrum participated in https://github.com/rust-lang/wg-async-foundations/issues/54 but I don't see his name on the list. (cc @Stu)


nikomatsakis:
Also, @rylev, I took a stab at adding folks who participated in the writing sessions, but I'm quite sure I'm missing lots of people! I've not been very good at keeping records. Maybe you want to toss in a few names?

Stupremee:
Oh that's weird. I will look into that!

PR #189, opened 2021-05-01
Shiny future: Performance analysis tools

farnz:
This is a first pass at describing a future with awesome perf tracing tools that we can use to identify and fix issues in code, when it either doesn't work at all (deadlock), or underperforms against expectations.

Feedback welcome!


src/vision/shiny_future/barbara_appreciates_performance_tools.md
```diff
@@ -0,0 +1,85 @@
...
+
+Fortunately, there are great tracing tools available for async Rust. Barbara turns on tracing, and immediately gets interesting information in her trace viewer. She can see all the tasks she has spawned, the lines of code where a `.await` returns control to the executor, and delays between a `Waker` being invoked and the corresponding `.await` resuming execution.
+
+With this information in hand, she finds a decompression path that is unexpectedly CPU-bound, because she can see a stack trace for the task that is running and blocking a woken up future from getting invoked again. The memory use of this future tells her that the compressed blobs are larger than she thought, but inspecting shows that this is reasonable. She thus puts the decompression onto its own blocking task, which doesn't fix things, but makes it clear that there is a deadlock passing data between two bounded channels; the trace shows the `Waker` for a `rx.next().await` being invoked, but the corresponding `.await` never runs. Looking into the code, she notes that the task is waiting on a `tx.send().await` call, and that the channel it is trying to send to is full.
```
pickfire, line 22: Wouldn't it be better if she could detect if a channel is full before that? I recall there are some articles saying they track channel usage as metrics but I think that maybe be too complicated for small teams.

src/vision/shiny_future/barbara_appreciates_performance_tools.md
```diff
@@ -0,0 +1,85 @@
...
+
+Alan is excited about how easy it is to find out when his projects don't work as expected. He's happy
+
+### **What is [Grace] most excited about in this future? Is she disappointed by anything?**
+
+Grace is happy because the performance tools give her all the low level insights she wants into her code, and shows her what's going on "behind the scenes" in the executor. As a C++ developer, she is also excited when she sees that Rust developers who see an issue with her services can give her useful information about exactly what they see her C++ doing - which she can correlate with her existing C++ performance tools via the unique ID.
+
+### **What is [Niklaus] most excited about in this future? Is he disappointed by anything?**
+
+Niklaus is content. The tooling tells him what he needs to know, and allows him to add interesting information to places where he'd otherwise be stuck trying to extract it via `println!()`. He's not entirely sure how to use some of the detailed information, but he can ignore it easily.
```
pickfire, line 50: Oh, that's what I did. But doesn't `println!()` cause some delays?

src/vision/shiny_future/barbara_appreciates_performance_tools.md
farnz: We can already do that, *if* we know that channels are going to fill up.

In this case, though, Barbara thought she had something emptying all the channels - but she didn't, because she has a classic ABBA deadlock.

src/vision/shiny_future/barbara_appreciates_performance_tools.md
farnz: Any debugging tool can cause some delays - the point of this "shiny future" story is to describe a case where the tools are amazing, and thus you can attach interesting information to traces, and filter them to see exactly what you need to know.

nikomatsakis:
Nice!

PR #191, opened 2021-05-05
Status quo barbara tries unix socket

### LINKED ISSUE #190: Implicit connection or not ready
### Brief summary

Recently I worked on low-level rust async libraries, I wanted to do a simple unix stream client but after tried it out first hand as I don't usually do low-level networking in rust, I faced something which is not expected, mainly around extra connection struct and missing `poll_fn` functions.

Since I just faced this just now, I will be writing a draft on this at least it is still fresh in my mind.

### Optional details

* (Optional) Which [character(s)] would be the best fit and why?
    * [ ] [Alan]: the experienced "GC'd language" developer, new to Rust
    * [ ] [Grace]: the systems programming expert, new to Rust
    * [ ] [Niklaus]: new programmer from an unconventional background
    * [x] [Barbara]: the experienced Rust developer
* (Optional) Which [project(s)] would be the best fit and why?
    * Sending a/multiple HTTP request using unix stream.
* (Optional) What are the key points or morals to emphasize?
    * The types in low-level rust networking (talking about hyper but could happen to others) are quite implicit, the users have to carefully check whether polling manually is required and check each output type one by one to make sure it will work.

[character(s)]: https://rust-lang.github.io/wg-async-foundations/vision/characters.html
[project(s)]: https://rust-lang.github.io/wg-async-foundations/vision/projects.html
[Alan]: https://rust-lang.github.io/wg-async-foundations/vision/characters/alan.html
[Grace]: https://rust-lang.github.io/wg-async-foundations/vision/characters/grace.html
[Niklaus]: https://rust-lang.github.io/wg-async-foundations/vision/characters/niklaus.html
[Barbara]: https://rust-lang.github.io/wg-async-foundations/vision/characters/barbara.html

pickfire:
Fix #190

Rendered https://github.com/pickfire/wg-async-foundations/blob/ready/src/vision/status_quo/barbara_tries_unix_socket.md

cc @Stupremee

### DIFF
--- a/src/vision/status_quo/barbara_tries_unix_socket.md
+++ b/src/vision/status_quo/barbara_tries_unix_socket.md
@@ -1,22 +1,5 @@
 # 😱 Status quo stories: Barbara tries Unix socket
 
-*This is a template for adding new "status quo" stories. To propose a new status quo PR, do the following:*
-
-* *Create a new file in the [`status_quo`] directory named something like `Alan_tries_to_foo.md` or `Grace_does_bar.md`, and start from [the raw source from this template]. You can replace all the italicized stuff. :)*
-* *Do not add a link to your story to the [`SUMMARY.md`] file; we'll do it after merging, otherwise there will be too many conflicts.*
-
-*For more detailed instructions, see the [How To Vision: Status Quo] page!*
-
-*If you're looking for ideas of what to write about, take a look at the [open issues]. You can also [open an issue of your own] to throw out an idea for others.*
-
-[How To Vision: Status Quo]: ../how_to_vision/status_quo.md
-[the raw source from this template]: https://raw.githubusercontent.com/rust-lang/wg-async-foundations/master/src/vision/status_quo/template.md
-[`status_quo`]: https://github.com/rust-lang/wg-async-foundations/tree/master/src/vision/status_quo
-[`SUMMARY.md`]: https://github.com/rust-lang/wg-async-foundations/blob/master/src/SUMMARY.md
-[open issues]: https://github.com/rust-lang/wg-async-foundations/issues?q=is%3Aopen+is%3Aissue+label%3Astatus-quo-story-ideas
-[open an issue of your own]: https://github.com/rust-lang/wg-async-foundations/issues/new?assignees=&labels=good+first+issue%2C+help+wanted%2C+status-quo-story-ideas&template=-status-quo--story-issue.md&title=
-
-
 ## 🚧 Warning: Draft status 🚧
 
 This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
@@ -27,6 +10,9 @@ If you would like to expand on this story, or adjust the answers to the FAQ, fee
 
 Content of `Cargo.toml` for reproducibility:
 
+<details>
+  <summary><code>Cargo.toml</code></summary>
+
 ```toml
 futures = "0.3.14"
 hyper = { version = "0.14.7", features = ["full"] }
@@ -34,8 +20,12 @@ pretty_env_logger = "0.4.0"
 reqwest = "0.11.3"
 tokio = { version = "1.5.0", features = ["macros", "rt-multi-thread"] }
 ```
+</details>
 
 There is a HTTP server in hyper which Barbara have to query.
+
+<details>
+  <summary>Server code</summary>
 
 ```rust
 use hyper::server::conn::Http;
@@ -66,8 +56,8 @@ async fn serve(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
         .unwrap();
     Ok(res)
 }
-
 ```
+</details>
 
 ## Nice simple query with high-level reqwest
 
@@ -92,6 +82,29 @@ Barbara search through reqwest doc and github issues to see how to use unix sock
 
 ## The screen stares at Barbara
 
+```rust
+use hyper::{body::HttpBody, client::conn, Body, Request};
+use tokio::net::UnixStream;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    pretty_env_logger::init();
+    let stream = UnixStream::connect("/tmp/socket").await?;
+
+    let (mut request_sender, connection) = conn::handshake(stream).await?;
+ 
+    let request = Request::get("/").body(Body::empty())?;
+    let mut response = request_sender.send_request(request).await?;
+    println!("{:?}", response.body_mut().data().await);
+ 
+    let request = Request::get("/").body(Body::empty())?;
+    let mut response = request_sender.send_request(request).await?;
+    println!("{:?}", response.body_mut().data().await);
+ 
+    Ok(())
+}
+```
+
 Barbara wrote some code according to the comments Barbara saw and read some docs like what is `handshake` to roughly know what it does. Barbara compile and it shows a warning, the `connection` variable is not used:
 ```
 warning: unused variable: `connection`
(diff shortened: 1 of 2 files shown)


src/vision/status_quo/barbara_tries_unix_socket.md
```diff
@@ -0,0 +1,314 @@
...
+futures = "0.3.14"
+hyper = { version = "0.14.7", features = ["full"] }
+pretty_env_logger = "0.4.0"
+reqwest = "0.11.3"
+tokio = { version = "1.5.0", features = ["macros", "rt-multi-thread"] }
+```
+
+There is a HTTP server in hyper which Barbara have to query.
+
+```rust
```
eminence, line 40: This block of code is just showing an example HTTP server, right?   And isn't what Barbara has written?  If so, maybe we can omit this (or at least hide it behind a `<details>` tag)

src/vision/status_quo/barbara_tries_unix_socket.md
```diff
@@ -0,0 +1,314 @@
...
+    tokio::spawn(async move {
+        if let Err(e) = connection.await {
+            eprintln!("error: {}", e);
+        }
+    })
+    
+    let request = ...
+```
```
eminence, line 123: Spawning a new task looks a little strange at first, but this is indeed what the hyper docs recommend.  Is that where you/Barbara got the idea to await the connection in this way?  If so, it would probably be good to mention that.

src/vision/status_quo/barbara_tries_unix_socket.md
```diff
@@ -0,0 +1,314 @@
...
+```rust
+ TRACE mio::poll > registering event source with poller: token=Token(0), interests=READABLE | WRITABLE
+Some(Ok(b"Hello World!"))
+ TRACE want      > signal: Want
+ TRACE mio::poll > deregistering event source from poller
+ TRACE want      > signal: Closed
+Error: hyper::Error(Canceled, "connection was not ready")
+```
+
+As a typical method of solving asynchronous issue. Barbara add prints to every await boundaries in the source code to understand what is going on.
```
eminence, line 142: I like this part (somethings the only thing you know how to do is println debugging).

src/vision/status_quo/barbara_tries_unix_socket.md
```diff
@@ -0,0 +1,314 @@
...
+
+Barbara still have no idea why it needs to be done this way. But at least it works now.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+### **What are the morals of the story?**
+
+Barbara is not able to see the problem right away. Usually missing an `await` is an issue but in this case, not awaiting on another variable or not polling for ready when using a low-level library may the program incorrect, it is also hard to debug and figure out what is the correct solution.
```
eminence, line 293: For me personally, this "solution" (adding the call to `future::poll_fn`) feels very unsatisfying!  Do you think Barbara felt the same way?  Did she feel like this was really the solution to her problem?  Or did it feel more like some workaround/bandaid?

eminence (changes requested):
Thanks for the story!  I was able to get a good sense for what Barbara was struggling through with this story.

src/vision/status_quo/barbara_tries_unix_socket.md
pickfire: Okay, good idea. Maybe we can also hide the `Cargo.toml` within `<details>`.

src/vision/status_quo/barbara_tries_unix_socket.md
```diff
@@ -0,0 +1,314 @@
...
+Barbara wrote some code according to the comments Barbara saw and read some docs like what is `handshake` to roughly know what it does. Barbara compile and it shows a warning, the `connection` variable is not used:
+```
+warning: unused variable: `connection`
+ --> src/main.rs:9:30
+  |
+9 |     let (mut request_sender, connection) = conn::handshake(stream).await?;
+  |                              ^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_connection`
+  |
+  = note: `#[warn(unused_variables)]` on by default
+```
```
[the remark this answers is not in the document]
pickfire: Nice, I think I missed this part, let me add it back.

src/vision/status_quo/barbara_tries_unix_socket.md
pickfire: I didn't know it is required to do it this way, but it does seemed weird. I didn't found the relevant part in docs until someone point me to it which is in the parent module. I got the idea until someone told me that it is required.

src/vision/status_quo/barbara_tries_unix_socket.md
pickfire: Yeah, feels like a magic to me, I wonder why it wasn't mentioned or like this is so implicit. It makes doing `async` no difference than `unsafe`, where I had to manually audit which block if they have completed. Barbara would probably feel the same way if she had no idea it is required in the first place. At least this part I think could be even marked `unsafe` since it is sort of "undefined behavior" in async world (not what the original spec says). At least to me I think this is understandable for a low-level library but it would be better if it is documented or marked as unsafe, at least users won't be confused, not sure if this can be embedded in the type system so it can be safe.

src/vision/status_quo/barbara_tries_unix_socket.md
```diff
@@ -0,0 +1,329 @@
...
+    println!("sent 2");
+    println!("{:?}", response.body_mut().data().await);
+```
+
+Barbara still have no idea why it needs to be done this way. But at least it works now.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
```
nikomatsakis, line 305:
```suggestion
```

src/vision/status_quo/barbara_tries_unix_socket.md
nikomatsakis: I would like to see some of *this* in the morals! "Feels like magic" and having to manually audit. That exact analogy to unsafe has occurred to me as well, reading some of the status quo stories.

nikomatsakis (changes requested):
Thanks @pickfire! I left a few suggestions, but I think this is basically ready to land.

src/vision/status_quo/barbara_tries_unix_socket.md
```diff
@@ -0,0 +1,329 @@
...
+warning: unused variable: `connection`
+ --> src/main.rs:9:30
+  |
+9 |     let (mut request_sender, connection) = conn::handshake(stream).await?;
+  |                              ^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_connection`
+  |
+  = note: `#[warn(unused_variables)]` on by default
+```
+
+Barbara then runs the program. Barbara was stares at the screen and the screen stares at her. Barbara waited and ... it was stuck. So Barbara decides to look at the logs and run it again with `env RUST_LOG=trace cargo r`, and it was indeed stuck, but not sure where.
```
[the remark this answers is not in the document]
pickfire: Oh, thanks for pinging this part, I saw a typo.

src/vision/status_quo/barbara_tries_unix_socket.md
pickfire:
```suggestion
Barbara is not able to see the problem right away. Usually missing an `await` is an issue but in this case, not awaiting on another variable or not polling for ready when using a low-level library may the program incorrect, it is also hard to debug and figure out what is the correct solution.

In a way, some of the fixes "feels like magic". Sometimes polling is required to be done but where? It may make people afraid of using `async/.await` and end up writing safety net code (for example, writing code to do type checking in weakly typed languages in every lines of code to be safe).

Having these pitfalls in mind, one can easily relate it back to unsafe. If there are unsafe blocks, the user needs to manually audit every specific code block for undefined behaviors. But in the case of async, the situation is someone similar such that the user need to audit the whole async code blocks (which is a lot compared to unsafe) for "undefined behaviors", rather than having when it compiles it works sort of behavior.
```

nikomatsakis (approved):

PR #192, opened 2021-05-06
Barbara async tracing

erichgess:
Shiny future story about provide tooling to trace events as they move through a chain or graph of async expressions.

There was no specific user story that this feature idea was based upon so I used some of my own experiences to create a story.  As such, would be grateful for any ideas to add some details or if anyone as specific experiences from Rust async which would have benefit from a tracing tool.


nikomatsakis (changes requested):
This reads like more like a status quo than a shiny future! I think it'd be great to expand that final paragraph to write the story 'as if' this dream feature already existed, and show how it worked (we might want to take the rest of the story and make it into a status quo story, though...)

erichgess:
> This reads like more like a status quo than a shiny future! I think it'd be great to expand that final paragraph to write the story 'as if' this dream feature already existed, and show how it worked (we might want to take the rest of the story and make it into a status quo story, though...)

Good point, will update tonight.

erichgess:
Here's a quick outline of how I'll update this to be a shiny future rather than status quo:

```none
Barbara is currently working on a project which makes heavy usage of async expressions.

After making implementing a new feature for the project she receives some feedback from users.

For a small number of users, they now experience periodic slow performance when before there were no issues

Obviously, Barbara suspects that the new changes are the culprit but she wants to verify her hypothesis quickly and with minimal work.

Fortunately, she has async tracing.  Which will let her trace events as they move through the flow of async expressions.

Barbara uses information from the users to recreate the slow issues on her testing environment and then turns on async tracing

She runs a few requests until one of them recreates the slow down issue and she looks at her trace.
 With the trace she can see each async expression that the request passed through and how long each expression took.

She quickly sees that the slow expression is occurring in a database call, but curiously, it’s code that existed prior to her changes.

Now she knows at least one place in the code that the slow down occurs.  She runs several more tests and verifies that it is always this database call.

She adds some additional logs and instrumentation and discovers that the database call is now running slowly because of a locked table.

She again goes back to the code she just added.  This problem did not occur before so it seems very likely that the new code is the root cause. She looks through her code and sees that the new code makes a database query that could result in a full table lock for a specific subset of customers.

Barbara rewrites her query so that it is no longer an issue.  Reruns her tests and confirms the fix.

Thanks to the async tracing feature, Barbara was able to quickly determine where the issue lay and fix it.
```

src/vision/shiny_future/barbara_wants_async_tracing.md
```diff
@@ -0,0 +1,106 @@
...
+actually lagged.  Each request experienced a slow down in different async expressions, but each expression
+had one thing in common they each queried the same database table. She also noticed that there was a relation
+in when the latency occurred: all the laggy requests tended to occur in clusters. From this she was able to identify
+that the root cause was some updates made to the database which led to some queries, if they arrived together,
+to run relatively slowly. With tracing, Barbara was saved the effort of having to meticulous work through the code
+and try to deduce what the cause was and she didn’t have to add in a large amount of logging or other
+instrumentation.  All the instrumentation and analysis was provided out of the box and required no development
+work for Barbara to isolate the cause. 
+
+Barbara can’t believe how much time she saved having this debugging tool provided out of the box.
```
nikomatsakis, line 42: This is awesome!

src/vision/shiny_future/barbara_wants_async_tracing.md
```diff
@@ -0,0 +1,106 @@
...
+some features.  The lag events are rare but Barbara and her team are determined to fix them.  With some work
+Barbara is able to recreate the lag reliably in the QA environment; but now she must figure out where in the
+complex code base this lag could be coming from and why it’s happening.  Fortunately, Rust’s `async` framework
+now provides a built in Tracing tool.  By building her service with the `tracing` flag on, her code is automatically
+instrumented and will start logging trace data to a file for later analysis.
+
+Barbara runs the instrumented code in QA and recreates the laggy event several times.  Then she takes the 
+generated trace file and looks through the data.  She immediately sees where each of the slow requests
+actually lagged.  Each request experienced a slow down in different async expressions, but each expression
+had one thing in common they each queried the same database table. She also noticed that there was a relation
```
nikomatsakis, line 34:
```suggestion
had one thing in common: they each queried the same database table. She also noticed that there was a relation
```

nikomatsakis (changes requested):
I love this story, thank you for converting it into more of a "future form". That said, I'm wondering if we can add more detail too -- I don't quite understand for example what kind of queries she is able to make. I'm not so worried about how the data is gathered, but I'd like a more concrete idea of what she is able to get out from the trace data (at least if you have something in mind).

erichgess:
> I love this story, thank you for converting it into more of a "future form". That said, I'm wondering if we can add more detail too -- I don't quite understand for example what kind of queries she is able to make. I'm not so worried about how the data is gathered, but I'd like a more concrete idea of what she is able to get out from the trace data (at least if you have something in mind).

I'll be honest, I just made this up this morning because I couldn't come up with a good situation for this story and I didn't want to get blocked. But now I've got a concrete example from my memory that I can put here and that will make this feel much more realistic.

nikomatsakis (approved):

PR #193, opened 2021-05-07
add "life of an aws engineer" status quo story

nikomatsakis:
This is a pretty long story, so I broke it into parts!

[Rendered Preview](https://nikomatsakis.github.io/wg-async-foundations/vision/status_quo/aws_engineer.html)


src/vision/status_quo/aws_engineer/figuring_out_the_best_option.md
```diff
@@ -0,0 +1,3 @@
+# Status quo of an AWS engineer: Figuring out the best option
+
+Sometimes after working on `AsyncRead`, Alan stumbles over the `async-trait` crate. This crate offers a macro that will let him add `async fn` to traits. He's excited about this because it seems like it would allow him to rewrite some of the custom `AsyncRead` impls in a cleaner way. The only problem is that he can't really judge what the implications are going to be -- will it be faster? Slower? It's hard to tell until it's done. He feels like this comes up a lot in Rust: he is forced to make a choice and see it all the way through to the end before he can decide whether he likes it (or if it will work at all: sometimes he encounters a compiler error part of the way through that he just can't figure out how to resolve). It's particularly frustrating in Async Rust where [there seem to be so many options to choose from](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_needs_async_helpers.html).
```
eminence, line 3:
```suggestion
Sometime after working on `AsyncRead`, Alan stumbles over the `async-trait` crate. This crate offers a macro that will let him add `async fn` to traits. He's excited about this because it seems like it would allow him to rewrite some of the custom `AsyncRead` impls in a cleaner way. The only problem is that he can't really judge what the implications are going to be -- will it be faster? Slower? It's hard to tell until it's done. He feels like this comes up a lot in Rust: he is forced to make a choice and see it all the way through to the end before he can decide whether he likes it (or if it will work at all: sometimes he encounters a compiler error part of the way through that he just can't figure out how to resolve). It's particularly frustrating in Async Rust where [there seem to be so many options to choose from](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_needs_async_helpers.html).
```

src/vision/status_quo/aws_engineer/missed_waker_leads_to_lost_performance.md
```diff
@@ -0,0 +1,12 @@
+# Status quo of an AWS engineer: Missed `Waker` leads to lost performance
+
+Once the server is working, Alan starts to benchmark it. He is not really sure what to expect, but he is hoping to see an improvement in performance relative to the baseline service they were using before. To his surprise, it seems to be running slower!
+
+Digging into the problem a bit, he wishes -- not for the first time -- that he had better tools to understand what was happening. This time he doesn't even bother trying to fire up the debugger. He just starts looking at his metrics. He's accumulated a pretty decent set of metrics by now, and he can often get a picture of the state of the runtime from them.
+
+After a few days of poking at the problem, Alan notices something odd. It seems like there is often a fairly large delay between the completion of a particular event and the execution of the code that is meant to respond to that event. Looking more closely, he realizes that the code for handling that event fails to trigger the `Waker` associated with the future, and hence the future never wakes up.
```
eminence, line 7: It would be interesting to read how Alan's poking eventually lead him to the root cause.  Was his set of metrics rich enough to discover this without using any other debugging techniques?  Or did he need to do something like println debugging with timestamps?

eminence (commented):
Just did a very quick pass, reading for typos and link issues.

I gotta say, I have a quite a bit of sympathy for Alan after reading this!

src/vision/status_quo/aws_engineer/getting_started_with_rust.md
```diff
@@ -0,0 +1,23 @@
+# Status quo of an AWS engineer: Getting started with Rust
+
+For his latest project, Alan is working on a new service, [DistriData]. Like many AWS services, they are trying to move on a tight deadline. 
+
+The plan is to implement the new service in Rust. The service that they are rewriting was implemented in Java, but it was having difficulty with high tail latencies and other performance hiccups, and they would like to reduce resource usage by adopting Rust.
```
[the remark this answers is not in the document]
nikomatsakis: Good question! I think replacing (and, often rearchitecting) parts of an existing service is much more common.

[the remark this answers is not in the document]
eminence: Makes sense.  Actually as I read the rest of the story, this became clear (since Alan eventually compared the performance of the new rust service to the old java service)

Maybe something like:

```suggestion
The plan is to implement the service in Rust.  This service is actually a rewrite of a service that is implemented in Java, but it is having difficulty with high tail latencies and other performance hiccups, and they would like to reduce resource usage by adopting Rust.
```

src/vision/status_quo/aws_engineer/missed_waker_leads_to_lost_performance.md
```diff
@@ -0,0 +1,12 @@
+# Status quo of an AWS engineer: Missed `Waker` leads to lost performance
+
+Once the server is working, Alan starts to benchmark it. He is not really sure what to expect, but he is hoping to see an improvement in performance relative to the baseline service they were using before. To his surprise, it seems to be running slower!
+
+Digging into the problem a bit, he wishes -- not for the first time -- that he had better tools to understand what was happening. This time he doesn't even bother trying to fire up the debugger. He just starts looking at his metrics. He's accumulated a pretty decent set of metrics by now, and he can often get a picture of the state of the runtime from them.
```
wilfredwee, line 5: Apologies if this should be a PR instead of a comment. Let me know and I can perhaps submit one tonight.

It's not clear how does Alan just start looking at his metrics. Here's how I imagine the story goes:

After trying a few common tricks to improve performance without avail, Alan decides to add more metrics and logs in his service, to understand where the bottlenecks are. Alan is used to using a well-supported internal tool (or a mature open source project) to collect metrics, where all he needed to do was pull in the library and set up a few configuration parameters.

However, in Rust, there is no widely-used, battle-tested library inside and outside his company. Even less so in an async code base! So Alan just used what seemed to be the best options: `tracing` and `metrics` crate, but he quickly found that they couldn't do a few of the things he wants to do, and somehow invoking the metrics is causing his service to be even slower. Now, Alan has to debug and profile his metrics implementation before he can even fix his service. (Cue another story on how that's difficult...)

nikomatsakis: Thanks!

src/vision/status_quo/aws_engineer/getting_started_with_rust.md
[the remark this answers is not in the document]
nikomatsakis: I made some tweaks.

src/vision/status_quo/aws_engineer/missed_waker_leads_to_lost_performance.md
nikomatsakis: added

PR #194, opened 2021-05-07
Add life of a Facebook C++ engineer status quo story

metajack:
This is currently an outline, similar in style to @nikomatsakis's life of an AWS engineer story.


metajack:
cc @pcwalton @bmwill who may have some feedback and suggestions

eminence:
The outline for the story seems pretty sensible in terms of what Grace would actually need to do, but it all seems to be generic rust stuff.  Where in this outline do you see async challenges coming into this story?

nikomatsakis:
I'm interested in both sync and async, whatever is effect peopling trying to build systems on async Rust.

nikomatsakis:
I'm going to merge, we can always add more details in subsequent PRs.

PR #195, opened 2021-05-08
Shiny Future: Buffered Async Read/Write

erichgess:
Shiny future story where there is a common abstraction for buffered `AsyncRead`/`AsyncWrite` that is implemented by all `async` runtimes.

### DIFF
--- a/src/vision/shiny_future/barbara_wants_async_rw.md
+++ b/src/vision/shiny_future/barbara_wants_async_rw.md
@@ -1,0 +1,83 @@
+# ✨ Barbara Wants Async Read Write
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "shiny future" story submitted as part of the brainstorming period. It is derived from what actual Rust users wish async Rust should be, and is meant to deal with some of the challenges that Async Rust programmers face today.
+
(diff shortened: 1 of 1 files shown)


eminence:
This feels a bit like a "status quo" story instead of a "shiny future".    Does it make sense to re-frame this story so that it assumes that the traits that Barbara are looking for are available, and then see how that affects Barbara's experience writing her `sans-io` library?

eminence:
I don't want to minimize Barbara's past experiences with AsyncRead and AsyncWrite, though.  Are there any status-quo stories that you think cover those experiences?  If so, those could be added as a link, maybe something like:

> The last time Barbara tried to write a runtime agnostic library, she ran into a lot of problems.  For example, there was no generic buffered AsyncRead trait to use.  [insert link to status quo story here]


Probably [Barbara writes a runtime-agnostic library](https://github.com/rust-lang/wg-async-foundations/blob/master/src/vision/status_quo/barbara_writes_a_runtime_agnostic_lib.md) is relevant here.  I just skimmed this, and I don't see any mention about buffered IO (so maybe that could be a good addition to that story)

erichgess:
My biggest concern is that I don't know if this story will have much heft to it when written with Buffered IO already being a part of the async common interface.  It seems like the story will simply be "Barbara wanted to write `sans-io` for Rust, but the performance wasn't as good as she'd hoped so she switched to Buffered IO"?  Is that sufficient for a Shiny Future story?

Maybe having a paragraph detailing her experience before Buffered IO?  Or that Buffered IO just got added to async and now Barbara is excited because she can add it to `sans-io` and maintain her design requirements and get the performance improvement?

PR #205, opened 2021-05-15
Added story about Alan writing a prefetch iterator

YohDeadfall:
[Rendered](https://github.com/rust-lang/wg-async-foundations/blob/7614fb569f62591601530b747657e6ab86529b89/src/vision/status_quo/alan_wants_prefetch_iterator.md)

### DIFF
--- a/src/vision/status_quo/alan_wants_prefetch_iterator.md
+++ b/src/vision/status_quo/alan_wants_prefetch_iterator.md
@@ -1,0 +1,42 @@
+# 😱 Status quo stories: Alan wants an async iterator with prefetch
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+Alan once wrote a data processing microservice in a GC'd language which was designed for high throughput. Now he wants to write it in Rust and have strong ownership model.
+
+The original service consumes messages from a source stream (e.g. Kafka), process them and produces results to another stream and/or saves them to a database. Since the service acquaries some data from other sources like external services and its own PostgreSQL database, Alan batches incoming messages to acquarie as much as possible data from that sources with minimal overhead.
+
+Since messages might arrive with some delays between them, or can end at some point for a while, their number is unknown, there's an async iterator which reads the input stream and waits some time before producing a batch if the next message isn't immediately ready.
+
+While this kind of iterator returns control to the caller, and the caller continues execution, there can be a running future preparing the next batch to be taken synchronously with a very high probability.
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+### **What are the morals of the story?**
+* Async looks not so hard after reading documentation and block posts on the topic, but when it comes to implementing your own async iterator, things stops seeming so beautiful as before. There's a very steep learning curve.
+
+### **What are the sources for this story?**
+Personal experience of the author.
+
+### **Why did you choose *NAME* to tell this story?**
+As a backend developer in a GC'd language, Alan writes async code every day. He wants to gain the maximum performance and have memory safety at the same time.
+
+### **How would this story have played out differently for the other characters?**
+*In some cases, there are problems that only occur for people from specific backgrounds, or which play out differently. This question can be used to highlight that.*
+
+[character]: ../characters.md
+[status quo stories]: ./status_quo.md
+[Alan]: ../characters/alan.md
+[Grace]: ../characters/grace.md
+[Niklaus]: ../characters/niklaus.md
+[Barbara]: ../characters/barbara.md
+[htvsq]: ../how_to_vision/status_quo.md
+[cannot be wrong]: ../how_to_vision/comment.md#comment-to-understand-or-improve-not-to-negate-or-dissuade
--- a/src/vision/status_quo/alan_wants_prefetch_iterator.md
+++ b/src/vision/status_quo/alan_wants_prefetch_iterator.md
@@ -57,7 +57,7 @@ let result_b = featrure_b.await;
 let result_c = featrure_c.await;
 
 // How it could be
-let (result_a, result_b, result_c) = join!(feature_a, feature_b, feature_c);
+let (result_a, result_b, result_c) = join!(feature_a, feature_b, feature_c).await;
 ```
 
 ## 🤔 Frequently Asked Questions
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Hmm, reading this story, I got a bit confused -- I couldn't quite tell what *happened*. Was Alan able to write the iterator or not? Also, I'm wondering if you've seen [Alan hates writing a stream](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/alan_hates_writing_a_stream.html), maybe that covers the material you are thinking of?

YohDeadfall:
He had not finished it at the moment of writing the story, but was successful with the first part because heard about futures before and, therefore, read blog posts on the topic to be prepared.

The problem here is two concurrently running futures inside. It happens when the inner stream returns poll and the current batch isn't filled yet. The second future here is a timer which wakes up the current iterator and which runs in parallel with the inner stream.

Alan thinks `std::sync::Mutex` can help in preventing the current iterator to be accessed by the timer and the inner stream simultaneously, so he can safely modify the state by adding another element to a batch, or store it in a field if the vector is full and wake up the caller. Therefore, looks like Alan will accomplish his tasks soon and even be able to determine was the current iterator triggered by the timer or not before acquiring the murex. It can help him to return the timer thread to the pool if he founds that the iterator already executes, and allow another operation to get that thread as soon as possible.

Now he wonders would it be possible to express that using generators as it was done in C#?

Alan explored `FeatureExt` from `async-std` and found no evidence that it's possible to wait for multiple features returning different results (it's not possible for `ValueTask`s in .NET, but it worked well with `Task`s which can be awaited multiple times). Having `await_all` and `await_any` might be a good addition.

`await_all` is similar to `join` and is a sugar over:

```rust
let f_a = do_async_a();
// ...
let f_z = do_async_z();

let result = (f_a.await, /*...,*/ f_z.await);
```

`await_any` should be like `race`, but returning  heterogeneous tuple.

YohDeadfall:
The most important thing here is that at least two threads run in parallel: the caller and the inner stream. Alan is at this point currently.

From his experience with a GC'd languages the prefetch iterator will be dropped only when no one references it and that iterator lives on the heap, but here, in Rust, the iterator will be put on the stack and be used by two threads simultaneously. So he thinks that drop should be protected by the murex too.

While it's not easy to implement a custom future or async iterator in Rust, Alan founds it more pleasant than doing the same without generators in a language to which he is used to. He wishes that the compiler can help him all the way down.

---

@nikomatsakis, I think that the moral of the story should changed to these points:

- The should be an easier way to write futures by procedural macros, syntactic sugar or maybe with some help from the IDE/analyzer side by tweaking the code and fixing issues.
- Basic macros and functions are needed to manipulate on multiple futures and give tools for task parallelization.

The "cliff" is already covered as you said, and while this story intersects with the previous this one goes beyond that as I see.

[As I read about `rayon`](https://developers.redhat.com/blog/2021/04/30/how-rust-makes-rayons-data-parallelism-magical), it enforces safety by ownership and borrowing. Maybe something can be done here, don't know. Alan and I are new to async in Rust, and have a lot of things to learn here.

nikomatsakis:
@YohDeadfall ok, I'm starting to see where you're going for -- basically, Alan has this *idea* he wants to realize, and he was able to do it in a GC'd language, but he is kind of just *stuck* figure out how to structure it in Rust?

nikomatsakis:
> found no evidence that it's possible to wait for multiple features returning different results

FWIW, you can do this with `race`, but you need to do some adaptation. Something like this:

```rust
enum Choices<A, B, C> {
    A(A),
    B(B),
    C(C),
}

let future_a = async move { A(future_a.await) };
let future_b = async move { B(future_b.await) };
let future_c = async move { C(future_c.await) };
match future_a.race(future_b).race(future_c).await {
    A(a) => ...,
    B(b) => ....,
    C(c) => ...,
}
```

YohDeadfall:
> Alan has this idea he wants to realize, and he was able to do it in a GC'd language, but he is kind of just stuck figure out how to structure it in Rust?

Exactly, some things are not so obvious for him and he is used to a good support for async programming and rich enough async library in a GC'd language he uses on daily basis.

---

I added a new part to the story to clarify Alan struggles we discussed. The problem should be better explained now.

src/vision/status_quo/alan_wants_prefetch_iterator.md
```diff
@@ -14,7 +14,51 @@ The original service consumes messages from a source stream (e.g. Kafka), proces
...
+
+// await the race:
+match future_a.race(future_b).race(future_c).await {
+    A(a) => ...,
+    B(b) => ....,
+    C(c) => ...,
+}
+```
+
+While that helped Alan, it was completely unobvius to him. He expected to see a macro accepting futures and producing a new future to be awaited:
```
nikomatsakis, line 39:
```suggestion
While that helped Alan, it was completely unobvious to him. He expected to see a macro accepting futures and producing a new future to be awaited:
```

src/vision/status_quo/alan_wants_prefetch_iterator.md
```diff
@@ -14,7 +14,51 @@ The original service consumes messages from a source stream (e.g. Kafka), proces
...
+
+```rust
+match race!(feature_a, feature_b, feature_c).await {
+    // ...
+}
+```
+
+Having `join!` would be nice too for Alan, so he can avoid binding variables to features which later shall be awaited:
```
nikomatsakis, line 47: FWIW I think [this exists](https://docs.rs/futures/0.3.15/futures/macro.join.html), but it seems like the point is that it is quite undiscoverable.

src/vision/status_quo/alan_wants_prefetch_iterator.md
```diff
@@ -0,0 +1,86 @@
...
+// How it could be
+let (result_a, result_b, result_c) = join!(feature_a, feature_b, feature_c);
+```
+
+## 🤔 Frequently Asked Questions
+
+*Here are some standard FAQ to get you started. Feel free to add more!*
+
+### **What are the morals of the story?**
+* Async looks not so hard after reading documentation and block posts on the topic, but when it comes to implementing your own async iterator, things stops seeming so beautiful as before. There's a very steep learning curve.
```
nikomatsakis, line 68: Can we tighten up these morals? I like the point you raised on the thread:

* Even though Alan had experience writing async code in other languages, he had a hard time figuring out how to relatively simple things in Rust, like joining or racing on futures of different types.

nikomatsakis (changes requested):

src/vision/status_quo/alan_wants_prefetch_iterator.md
YohDeadfall: Good, I named the macro `join` only because saw a method with the same name and meaning. I think that lack of `await` here might bring some confusion since any feature is awaited and there`s a keyword for that you type each time.

src/vision/status_quo/alan_wants_prefetch_iterator.md
```diff
@@ -0,0 +1,86 @@
...
+let featrure_a = do_async_a();
+let featrure_b = do_async_b();
+let featrure_c = do_async_c();
+
+let result_a = featrure_a.await;
+let result_b = featrure_b.await;
+let result_c = featrure_c.await;
+
+// How it could be
+let (result_a, result_b, result_c) = join!(feature_a, feature_b, feature_c).await;
```
YohDeadfall, line 60: @nikomatsakis, I added `await` here, so now it really looks like async code and matches `race!` macro I proposed before. 6 characters longer to write, but it gives a lot of clarity in my opinion.

nikomatsakis:
Thanks. I'm going to merge!

PR #206, opened 2021-05-17
minor changes to Barbara's story

glommer:
One of the two issues identified in my call with @nikomatsakis is the time aspects that may be present in the decisions of when to build something new or improve the ecosystem.

Rather than writing a totally new story about it, I decided to add this as part of the motivation in Barbara's story.
The rest of her experience is very similar to mine.


glommer:
Let me know if this is acceptable!

nikomatsakis (approved):

nikomatsakis:
@glommer good idea!

Issue #150, opened 2021-04-14
GC'd language variant: Go


jsha:
Thanks for working on this project!

https://rust-lang.github.io/wg-async-foundations/vision/characters/alan.html

I think a "Variant D: Go" would be useful here. Proposed text:

Alan develops networking programs in Go. He enjoys the simplicity and first-class treatment of concurrency, and is excited about Rust's promise of "fearless concurrency." He'd like to try Rust for more efficient use of memory and CPU, as well as its ability to do FFI well. When working with async Rust, Alan has the question: why not use coroutines / green threads / M:N threads?

nikomatsakis:
I'd be happy to add such a variant, I would also love to hear if there are more specific stories that result from it.

PR #209, opened 2021-05-22
status_quo: add "variant Go", "alan builds a task scheduler"

gyuho:
For https://github.com/rust-lang/wg-async-foundations/issues/150.

Any feedback would be appreciated!

Thanks in advance.

### DIFF
--- a/src/vision/status_quo/alan_builds_a_task_scheduler.md
+++ b/src/vision/status_quo/alan_builds_a_task_scheduler.md
@@ -1,0 +1,704 @@
+# 😱 Status quo stories: Alan builds a task scheduler
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+A core component of [`DistriData`][distridata], called `TaskScheduler`, is in charge of (1) receiving client requests via an HTTP server, (2) serializing them in a task queue, (3) relaying each task to the state machine applier (e.g., apply change to the storage backend), and (4) returning the result back to the client.
+
+`TaskScheduler` was originally implemented in Go. New to Rust, [Alan][] believes Rust could provide the same quality of service but with less memory. Then decides to reimplement `TaskScheduler` in Rust, without knowing the challenges ahead.
+
+Alan only read the first few chapters of [Rust book](https://doc.rust-lang.org/nightly/book/title-page.html) to understand the core concepts like ownership model and syntax. Already proficient in Go, Alan jumped into the coding by working through a hands-on project. Alan often referred to the examples found in each Rust crate but may lack deep understanding of how Rust works. Alan first focused on translating the Go code to Rust and as a result, the first iteration may be filled with non-idiomatic Rust code.
+
+### **Implementing request ID generator**
+
+Alan first transliterates request ID generator code, originally written in Go:
+
+```go
+import "sync/atomic"
+
+type Generator interface {
+	next() uint64
+}
+
+type generator struct {
+	prefix uint64
+	suffix uint64
+}
+
+func (gen *generator) next() uint64 {
+	suffix := atomic.SwapUint64(&gen.suffix, gen.suffix+1)
+	id := gen.prefix | (suffix & (math.MaxUint64 >> 16))
+	return id
+}
+```
+
+Alan learns Rust trait as the closest concept to Go interface but is now torn between [std::sync::atomic](https://doc.rust-lang.org/std/sync/atomic) and [crossbeam::atomic::AtomicCell](https://docs.rs/crossbeam). Reading multiple articles about how great crossbeam is and for its thread-safety promises, Alan chooses crossbeam (see ["crates better than std (from Reddit)"](https://www.reddit.com/r/rust/comments/hat5bt/what_are_your_favorite_better_than_std_crates/)):
+
+```rust
+use crossbeam::atomic::AtomicCell;
+
+pub struct Generator {
+    prefix: u64,
+    suffix: AtomicCell<u64>,
+}
+
+impl Generator {
+    pub fn new(...) -> Self {
+        ...
+    }
+
+    pub fn next(&self) -> u64 {
+        let suffix = self.suffix.fetch_add(1);
+        let id = self.prefix | (suffix & (u64::MAX >> 16));
+        id
+    }
+}
+```
+
+Accustomed to an opinionated way of doing concurrency in Go, Alan loses confidence in Rust async support, as he sees fragmented but specialized solutions in Rust async ecosystem.
+
+### **Implementing event notifier**
+
+Alan then implements the notifier to propagate the request and apply the progress with the scheduler and low-level state machine. In Go, it can be simply implemented as below:
+
+```go
+type Notifier interface {
+	register(id uint64) (<-chan string, error)
+	trigger(id uint64, x string) error
+}
+
+type notifier struct {
+	mu       sync.RWMutex
+	requests map[uint64]chan string
+}
+
+func (ntf *notifier) register(id uint64) (<-chan string, error) {
+	ntf.mu.Lock()
+	defer ntf.mu.Unlock()
+	ch := ntf.requests[id]
+	if ch != nil {
+		return nil, fmt.Errorf("dup id %x", id)
+	}
+
+	ch = make(chan string, 1)
+	ntf.requests[id] = ch
+	return ch, nil
+}
+
+func (ntf *notifier) trigger(id uint64, x string) error {
+	ntf.mu.Lock()
+	ch, ok := ntf.requests[id]
+	if ch == nil || !ok {
+		ntf.mu.Unlock()
+		return fmt.Errorf("request ID %d not found", id)
+	}
+	delete(ntf.requests, id)
+	ntf.mu.Unlock()
+	ch <- x
+	close(ch)
+	return nil
+}
+```
+
+Alan now needs the equivalent to Go `sync.RWMutex`, and found multiple options:
+
+- [`std::sync::RwLock`](https://doc.rust-lang.org/std/sync/struct.RwLock.html)
+- [`parking_lot::RwLock`](https://docs.rs/parking_lot)
+
+Already losing confidence in Rust std, Alan instead chooses `parking_lot`, as it claims up to 5x faster performance than `std::sync::Mutex` (see [github](https://github.com/Amanieu/parking_lot#parking_lot)). After numeruous hours of trials and errors, Alan discovered that `parking_lot::RwLock` is not intended for async/future environments (see [github issue](https://github.com/Amanieu/parking_lot/issues/86)). Having to think about which library to use for thread and async programming, Alan appreciates the simplicity of Go concurrency where threads are effectively abstracted away from its users. Alan is now using [`async_std::sync::RwLock`](https://docs.rs/async-std/1.9.0/async_std/sync/struct.RwLock.html) which seems nicely integrated with Rust async programming.
+
+To send and receive events, Alan needs the equivalent of Go channel but is not sure about [`std::sync::mpsc::channel`](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html), as he sees two other options: [Flume](https://github.com/zesterer/flume) which claims to be much faster than std (see ["Flume, a 100% safe MPSC that's faster than std (from Reddit)"](https://www.reddit.com/r/rust/comments/fj17z6/flume_a_100_safe_mpsc_thats_faster_than_std_and/)), and [`crossbeam_channel`](https://docs.rs/crossbeam-channel/0.5.1/crossbeam_channel/). Having used crossbeam, Alan chose crossbeam channel:
+
+```rust
+use async_std::sync::RwLock;
+use crossbeam_channel::{self, unbounded};
+
+pub struct Notifier {
+    requests: RwLock<HashMap<u64, crossbeam_channel::Sender<String>>>,
+}
+
+impl Notifier {
+    pub fn new() -> Self {
+        Self {
+            requests: RwLock::new(HashMap::new()),
+        }
+    }
+
+    pub fn register(&self, id: u64) -> io::Result<crossbeam_channel::Receiver<String>> {
+        let mut _mu;
+        match self.requests.try_write() {
+            Some(guard) => _mu = guard,
+            None => return Err(...),
+        }
+
+        let (request_tx, request_rx) = unbounded();
+        if _mu.get(&id).is_none() {
+            _mu.insert(id, request_tx);
+        } else {
+            return Err(...)
+        }
+
+        Ok(request_rx)
+    }
+
+    pub fn trigger(&self, id: u64, x: String) -> io::Result<()> {
+        let mut _mu;
+        match self.requests.try_write() {
+            Some(guard) => _mu = guard,
+            None => return Err(...),
+        }
+
+        let request_tx;
+        match _mu.get(&id) {
+            Some(ch) => request_tx = ch,
+            None => return Err(...),
+        }
+
+        match request_tx.send(x) {
+            Ok(_) => _mu.remove(&id),
+            Err(e) => return Err(...),
+        }
+
+        Ok(())
+    }
+}
+```
+
+Alan is still not sure if `crossbeam_channel` is safe for async programming and whether he should instead use another crate [`async_std::channel`](https://docs.rs/async-std/1.9.0/async_std/channel/index.html). While `crossbeam_channel` seems to work, Alan is not confident about his choice. Disgruntled with seemingly unnecessary divergence in the community, Alan wonders why all those cool improvements had not been made back to Rust core std libraries.
+
+### **Implementing task applier**
+
+Alan implements a task applier, which simply echoes the requested message, as in Go:
+
+```go
+type EchoManager interface {
+	apply(req *EchoRequest) (string, error)
+}
+
+type echoManager struct {
+	mu sync.RWMutex
+}
+
+func (ea *echoManager) apply(req *EchoRequest) (string, error) {
+	ea.mu.Lock()
+	defer ea.mu.Unlock()
+	switch req.Kind {
+	case "create":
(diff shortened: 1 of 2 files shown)


src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+    pub fn new() -> Self {
+        Self {
+            requests: RwLock::new(HashMap::new()),
+        }
+    }
+
+    pub fn register(&self, id: u64) -> io::Result<crossbeam_channel::Receiver<String>> {
+        let mut _mu;
+        match self.requests.try_write() {
+            Some(guard) => _mu = guard,
```
pickfire, line 133: Why this seemed so weird? Like `_mu = guard`.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+            None => return Err(...),
+        }
+
+        let (request_tx, request_rx) = unbounded();
+        if _mu.get(&id).is_none() {
+            _mu.insert(id, request_tx);
```
pickfire, line 139: If this is used, why does it need to have underscore? Also, why is `_mu` which is a mutex guard named as `mutex`? Shouldn't it be `writer` or `requests`?

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+        }
+        match req.kind.as_str() {
+            "create" => Ok(format!(
+                "SUCCESS create {}",
+                to_string(req.message.to_owned())
+            )),
+            "delete" => Ok(format!(
+                "SUCCESS delete {}",
+                to_string(req.message.to_owned())
+            )),
```
pickfire, line 230: Stringly-typed API is not a good idea in rust, we should switch this to something else. Like `Create` and `Delete` within an enum.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+```
+
+The code compiles and thus must be safe. However, after discussing the code with [Barbara][], Alan learns that `std::sync::Mutex` protects data from concurrent access, but `std::sync::Mutex` itselt must be also protected between threads. And this is where [`std::sync::Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html) comes in to provide safe multi-threaded access to the `Mutex`. Deeply confused, Alan made a quick fix to wrap `Mutex` with `Arc`:
+
+```rust
+// 2nd version
+use async_std::{sync::Arc, sync::RwLock};
+
+pub struct Manager {
+    mu: Arc<RwLock<()>>,
```
pickfire, line 244: What is `mu`?

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+        }
+    }
+    ...
+```
+
+This raises multiple questions for Alan:
+
+1. If `Mutex` itself had to be protected, why `Arc` is not unified into a single type? Is the flexibility of having different types really worth the less safety guarantee?
+2. Rust claims unparalleled safety. Is it still true for async programming? Rust compiler not complaining about the missing `Arc` means `Mutex` is still safe without `Arc`?
+3. What happens if the code went into production without `Arc`? Would the code have race conditions?
```
pickfire, line 260: If rust does not complain then it should not have race condition, if you faced this you might want to give an example.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+4. Does having `Arc` make code slower? Did I just introduce extra runtime cost?
```
pickfire, line 261: The thing is it does not always require `Arc`, sometimes just `Rc` will do. But most of the async ecosystem put an assumption that async may be using multiple threads so it is almost always wrapped in `Arc` even when it is not necessary (like `actix-net` which uses single-thread). Yes, both `Arc` and `Rc` have a cost but `Arc` have a higher cost.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+        Self {
+            mu: Arc::new(RwLock::new(())),
+        }
+    }
+    ...
+```
+
+This raises multiple questions for Alan:
+
+1. If `Mutex` itself had to be protected, why `Arc` is not unified into a single type? Is the flexibility of having different types really worth the less safety guarantee?
```
pickfire, line 258: Technically, this is not a question in async. Both `Arc` and `Mutex` are solutions to different problems, just that in this case using both is required. I think this is also for sync and the book should have show some examples about these. I am not sure if there are any benefits unifying it into a single type.

Talking about that here feels off-topic to me at least, because mainly this comes from sync and I don't think there is any issue with this. Unless separating them is a perf issue, but otherwise others would have merged it already.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+2. Rust claims unparalleled safety. Is it still true for async programming? Rust compiler not complaining about the missing `Arc` means `Mutex` is still safe without `Arc`?
+3. What happens if the code went into production without `Arc`? Would the code have race conditions?
+4. Does having `Arc` make code slower? Did I just introduce extra runtime cost?
+5. Which one is safe for async programming: `std::sync::Arc` and `async_std::sync::Arc`?
```
pickfire, line 262: Nice, I think most people stumbled across this. Both is safe when rust complains. But the issue is **when**, I think people are usually confused here.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+
+
+pub async fn apply_async(
+    notifier: Arc<notify::Notifier>,
+    ...
+) -> io::Result<()> {
+  ...
+```
+
+Alan is satisfied with the compilation success for the moment, but doesn't feel confident about the production readiness of Rust async.
```
pickfire, line 450: I don't think this part is related to async rust. It's just how rust, ownership system works. It is the same in sync rust. And it is not related to production readiness, if you use rust concepts and port it directly to go, it will be roughly the same.

src/vision/status_quo/alan_builds_a_task_scheduler.md
gyuho: Please note this is an attempt to transliterate Go code to Rust, where it's common to define an empty `sync.Mutex` field in struct with a variable name `mu`. I will name it better for this example :)

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+		}
+		fmt.Fprint(w, s)
+
+	default:
+		http.Error(w, "Method Not Allowed", 405)
+	}
+}
+```
+
+For Rust, Alan is faced with multiple choices again:
```
pickfire, line 536: In this case I think rust is better here. I think web server handling is too complex to have a "standard" way of doing things. Having choices is better than being forced with a single choice.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+
+- [hyper](https://github.com/hyperium/hyper)
+- [actix-web](https://github.com/actix/actix-web)
+- [warp](https://github.com/seanmonstar/warp)
+- [rocket](https://github.com/SergioBenitez/Rocket)
+- [tide](https://github.com/http-rs/tide)
+
+Alan strongly believes in Go's minimal dependency approach, and thereby chooses "hyper" for its low-level API. While "hyper" is meant to be a low-level building block, implementing a simple request handler in "hyper" still requires four different external dependencies. Alan is not surprised anymore, and rather accepts the status quo of split Rust ecosystem:
```
pickfire, line 544: I am surprised by this since hyper would be my last choice. hyper already mentioned that they are "low-level" and therefore is not recommended for normal use. Did you faced the same situation in the past but still chose hyper? Can I know if having a minimum amount of dependency was the reason? But yes, I agree that there is no single official comparison or a guide to choosing rust frameworks is a bad thing, things would have been easier if there is sort of feature comparison and review between them to compare, cargo-crev might help but it does not provide comparison.

src/vision/status_quo/alan_builds_a_task_scheduler.md
pickfire: Ah, yes I noticed that. But I think a better approach would be to translate module by module based on functionality if you are translating from go to rust. Go mental model does not fit rust very well, if you are translating from python to rust then it may be a good idea to translate line by line but not quite a good idea for go. If you are translating, thinking in terms of how data flow in rust would ease the translation, like what data should a struct contains and such. But yeah, what you should could the the first version of translation. Maybe I am looking at the status quo differently, now it seemed like the hard part is the translation part from go to rust as the memory model is different.

pickfire:
@gyuho Thanks for sending the pull request. Yes, this may likely be someone experienced with go but new to rust, like Alan. Most of these issues arise that rust have a different mental memory model and how types are used to make application type safe, having a high-level API but giving low-level control which leads to its complexity. Essentially, forcing a good design and best practices to a new developer to ensure nothing is wrong later on (like no race conditions).

Yes, I do agree to those async frameworks split, but what you showed is just a glimpse, the worse parts are tokio vs async-std vs smol split which affects even the high-level frameworks. What was faced here are mainly due to complexity but at least there are no "unexpected" behaviors (like async footguns).

I wonder if having something similar to cargo-crev but review across crates on high-level API, targets, design and such would make someone new to rust make choices like these easier since there are lots of choices.

src/vision/status_quo/alan_builds_a_task_scheduler.md
anguslees: Arc has various memory barriers for cross-cpu synchronisation; Rc does not and thus can be reordered/optimised more aggressively, including not writing out to memory at all if the compiler can prove it's equivalent.  In the worst contended case, Arc is similar to a mutex - you're thrashing a cpu cache line between cpus.

nikomatsakis (commented):
I gave this a skim -- this is great, thanks! I'm inclined not to do too many edits at this point, since we're trying to reach closure on the vision doc, and I found that I could largely follow what was happening.

gyuho:
@anguslees made some improvements to my Rust code (see https://github.com/gyuho/task-scheduler-examples/pull/1). I am still new, so let me know if we need to put some disclaimer that the example is not the most idiomatic Rust code :)

src/vision/status_quo/alan_builds_a_task_scheduler.md
anguslees: Afaics async_std::sync::Arc _is_ std::sync::Arc.  They're the same symbol, just exposed under async_std for some reason: https://docs.rs/async-std/1.9.0/src/async_std/sync/mod.rs.html#177

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+A core component of [`DistriData`][distridata], called `TaskScheduler`, is in charge of (1) receiving client requests via an HTTP server, (2) serializing them in a task queue, (3) relaying each task to the applier, and (4) returning the result back to the client.
```
rylev, line 11: nit: I got hung up on the term "applier". Is that a well known term that I'm just not familiar with? If not, maybe replacing it with something a tiny bit more descriptive would help.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+
+    pub fn next(&self) -> u64 {
+        let suffix = self.suffix.fetch_add(1);
+        let id = self.prefix | (suffix & (u64::MAX >> 16));
+        id
+    }
+}
+```
+
+Accustomed to an opinionated way of doing concurrency in Go, Alan loses confidence in Rust async support, as he sees fragmented but specialized solutions in Rust async ecosystem.
```
rylev, line 61: This is interesting although it's not related to _async_ per say. Atomics are also heavily used in synchronous code. I think we should leave this here though.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+
+### **Implementing event notifier**
+
+Alan then implements notifier to propagate request apply progress with scheduler and low-level state machine. In Go, it can be simply implemented as below:
```
rylev, line 65: I'm not sure I understand what this sentence is saying. Can you rephrase this? I attempted adding what appeared to be some missing comas and "the"s, but I'm still not sure. 

```suggestion
Alan then implements the notifier to propagate the request and apply the progress with the scheduler and low-level state machine. In Go, it can be simply implemented as below:
```

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+                "SUCCESS delete {}",
+                to_string(req.message.to_owned())
+            )),
+            _ => Err(...),
+        }
+    }
+}
+```
+
+The code compiles and thus must be safe. However, after discussing the code with [Barbara][], Alan learns that `std::sync::Mutex` protects data from concurrent access, but `std::sync::Mutex` itselt must be also protected between threads. And this is where [`std::sync::Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html) comes in to provide safe multi-threaded access to the `Mutex`. Deeply confused, Alan made a quick fix to wrap `Mutex` with `Arc`:
```
rylev, line 237: > The code compiles and thus must be safe. However, after discussing the code with [Barbara][], Alan learns that `std::sync::Mutex` protects data from concurrent access, but `std::sync::Mutex` itselt must be also protected between threads.

Hmm this isn't really true, or at least it's phrased in a potentially confusing way. `Arc` must be used to make the type useable from several threads. The code will not compile if you try to use it from multiple threads. This is relevant, but perhaps there was a misunderstanding in the communication between Barbara and Alan?

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,704 @@
...
+            mu: Arc::new(RwLock::new(())),
+        }
+    }
+    ...
+```
+
+This raises multiple questions for Alan:
+
+1. If `Mutex` itself had to be protected, why `Arc` is not unified into a single type? Is the flexibility of having different types really worth the less safety guarantee?
+2. Rust claims unparalleled safety. Is it still true for async programming? Rust compiler not complaining about the missing `Arc` means `Mutex` is still safe without `Arc`?
```
rylev, line 259: Yes, the code will not compile once you try to use the type in more than thread.

src/vision/status_quo/alan_builds_a_task_scheduler.md
rylev: Please remember that these stories are meant to reflect lived experiences. It's not necessarily relevant whether having choice is better than having a standard way of doing things. If this caused Alan confusion or hesitation, it should be noted.

rylev (commented):
I think this is a good view into potential confusion when writing async Rust from an experienced Go developers POV. A lot of the details here aren't necessarily specific to async Rust (e.g., atomics, Arc, etc.), but it's still all relevant. 

I did see a few misunderstandings of basic Rust types (e.g., `Arc`). Perhaps it would be relevant to note how Alan learned Rust. Was this Alan's first project in Rust?

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,698 @@
...
+
+async fn handle_request(
+    addr: SocketAddr,
+    req: Request<Body>,
+    applier: Arc<apply::Applier>,
+) -> Result<Response<Body>, hyper::Error> {
+    let http_version = req.version();
+    let method = req.method().clone();
+    let cloned_uri = req.uri().clone();
+    let path = cloned_uri.path();
```
pickfire, line 608: It would be better if we don't need to clone this. We can just do `req.uri().path()` below.

Side tip, people will usually try to avoid `clone()` and `unwrap()` in rust, since it's not always necessary, but sometimes it will make life easier.

src/vision/status_quo/alan_builds_a_task_scheduler.md
```diff
@@ -0,0 +1,698 @@
...
+                            }
+                            if success {
+                                match applier.apply(req).await {
+                                    Ok(rs) => resp = Response::new(Body::from(rs)),
+                                    Err(e) => {
+                                        resp = Response::builder()
+                                            .status(StatusCode::INTERNAL_SERVER_ERROR)...
+                                    }
+                                }
+                            }
```
pickfire, line 652: This could be structured in a better way. Like we don't need to keep a separate variable for success.

gyuho:
> Maybe can you write down the experience that you have as Alan, like what kind of knowledge did you have in rust before writing this to let others know by not knowing what material can face this kind of issues. Like the follow.

Great callout. Given the sheer number of concepts in Rust (compared to other languages), Alan (me, personally) did not want to wait till the end of the book (which may take multiple months): The `Arc<Mutex>` is mentioned in the chapter 16 https://doc.rust-lang.org/book/ch16-03-shared-state.html. Yes, https://doc.rust-lang.org/std/sync/struct.Mutex.html explains `Arc` but if you are given with multiple mutex options and happen to choose `async_std`, there's no mention of `Arc` here https://docs.rs/async-std/0.99.7/async_std/sync/struct.RwLock.html. If Alan didn't have to worry about choosing between multiple mutex libraries and just went with std mutex, he would have better understanding by looking at the official doc. I believe this is pretty typical to other new comers as well, and contributed to Alan's lack of understanding on some Rust concepts. I added more details on the "The story" section.

nikomatsakis (approved):

nikomatsakis:
Hey @gyuho -- I merged this, not sure if you had further changes in mind, but if so, follow-up PRs welcome!

PR #210, opened 2021-05-24
Add src/vision/status_quo/barbara_polls_a_mutex.md

sadmac7000:
This is a rough vision about some complexities I ran into trying to poll async mutexes.

### DIFF
--- a/src/vision/status_quo/barbara_polls_a_mutex.md
+++ b/src/vision/status_quo/barbara_polls_a_mutex.md
@@ -1,0 +1,151 @@
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## Barbara polls a Mutex
+
+### Brief summary
+
+[Barbara] is implementing an interpreter for a scripting language. This
+language has implicit asynchronicity, so all values in the language can
+potentially be futures underneath.
+
+Barbara wants to store a namespace which maps variable names to their values. She
+chooses to use a `HashMap` and finds the `async_lock` crate provides an async
+mutex, which she can use for concurrency. She determines she'll need a lock
+around the namespace itself to protect concurrent modification.
+
+For the entries in her map, Barbara decides to implement a two-variant enum. One
+variant indicates that there is no implicit asynchronicity to resolve and the
+value is stored directly here. The other variant indicates that this value is
+being computed asynchronously and polling will be required to resolve it.
+Because an asynchronous task might want to change one of these entries from the
+asynchronous variant to the ready variant, she'll need to wrap the entries in
+an `Arc` and a `Mutex` to allow an asynchronous task to update them.
+
+Barbara wants to be able to derive a future from the entries in her namespace
+that will allow her to wait until the entry becomes ready and read the value.
+She decides to implement the `Future` trait directly. She's done this before
+for a few simple cases, and is somewhat comfortable with the idea, but she runs
+into significant trouble trying to deal with the mutex in the body of her poll
+function. Here are her attempts:
+
+```rust=
+use async_lock::Mutex;
+
+enum Value {
+    Int(i32),
+}
+
+enum NSEntry {
+    Ready(Value),
+    Waiting(Vec<Waker>),
+}
+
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/barbara_polls_a_mutex.md
```diff
@@ -0,0 +1,151 @@
...
+an `Arc` and a `Mutex` to allow an asynchronous task to update them.
+
+Barbara wants to be able to derive a future from the entries in her namespace
+that will allow her to wait until the entry becomes ready and read the value.
+She decides to implement the `Future` trait directly. She's done this before
+for a few simple cases, and is somewhat comfortable with the idea, but she runs
+into significant trouble trying to deal with the mutex in the body of her poll
+function. Here are her attempts:
+
+```rust=
```
rylev, line 35: nit: remove the "="

src/vision/status_quo/barbara_polls_a_mutex.md
```diff
@@ -0,0 +1,151 @@
...
+enum Value {
+    Int(i32),
+}
+
+enum NSEntry {
+    Ready(Value),
+    Waiting(Vec<Waker>),
+}
+
+type Namespace = Mutex<String, Arc<Mutex<NSEntry>>>;
```
rylev, line 47: Is this meant to be `HashMap` and not `Mutex`?

src/vision/status_quo/barbara_polls_a_mutex.md
```diff
@@ -0,0 +1,151 @@
...
+variant indicates that there is no implicit asynchronicity to resolve and the
+value is stored directly here. The other variant indicates that this value is
+being computed asynchronously and polling will be required to resolve it.
+Because an asynchronous task might want to change one of these entries from the
+asynchronous variant to the ready variant, she'll need to wrap the entries in
+an `Arc` and a `Mutex` to allow an asynchronous task to update them.
+
+Barbara wants to be able to derive a future from the entries in her namespace
+that will allow her to wait until the entry becomes ready and read the value.
+She decides to implement the `Future` trait directly. She's done this before
```
rylev, line 30: It's unclear to me why implementing `Future` directly is desirable. Can you talk a bit more about why Barbara did not use async/await notation?

src/vision/status_quo/barbara_polls_a_mutex.md
```diff
@@ -0,0 +1,151 @@
...
+
+// Attempt 1: This compiles!!
+struct NSValueFuture(Arc<Mutex<NSEntry>>);
+impl Future for NSValueFuture {
+    type Output = Value;
+    pub fn poll(
+        self: Pin<&mut Self>, 
+        cx: &mut Context<'_>
+    ) -> Poll<Self::Output> {
+        let entry = match self.0.lock().poll() {
```
rylev, line 57: So does NSEntry also implement future?

src/vision/status_quo/barbara_polls_a_mutex.md
```diff
@@ -0,0 +1,151 @@
...
+            Poll::Ready(ent) => ent,
+
+            // When this returns, it will drop the future created by lock(),
+            // which drops our position in the lock's queue.
```
rylev, line 61: Presumably `lock()` returns a guard which will not drop the entry it contains when the guard is dropped. Can you explain this a bit more?

src/vision/status_quo/barbara_polls_a_mutex.md
```diff
@@ -0,0 +1,151 @@
...
+        // Pulled out pin-project, got confused, decided to just use unsafe.
+        match unsafe { Pin::new_unchecked(&mut self).lock_fut.unwrap() }.poll(cx) {
+            ...
+        }
+        // ??? lifetime for MutexLockFuture ???
+        // try async-std, async-lock
+    }
+}
+
+// Realize `lock_arc()` is a thing
```
rylev, line 95: I'm lost here. Perhaps some more exposition can take us through Barbara's thought process?

rylev (commented):
I think more stories about implementing futures manually are a good idea, but I'm having a hard time following this one. Perhaps some exposition is missing after the code is introduced.

src/vision/status_quo/barbara_polls_a_mutex.md
tmandry: I seem to recall that the necessary property is that [`MutexGuardArc<T>`](https://docs.rs/async-std/1.9.0/async_std/sync/struct.MutexGuardArc.html) impls `Send` if `T` is `Send`, unlike other `MutexGuard` types.

tmandry:
@sadmac7000 – The story gets the point across that doing this is hard, but it would also benefit from the clarifications suggested by @rylev.

nikomatsakis:
OK, going to assume that @sadmac7000 hasn't had time to follow-up. I'm going to merge, but would welcome follow-up PRs adding clarifications!

PR #214, opened 2021-05-28
status quo: Alan missing C# async

Nashenas88:
[Rendered](https://github.com/Nashenas88/wg-async-foundations/blob/f2cc60f61da71a3826dae686181845ba82662210/src/vision/status_quo/alan_misses_c_sharp_async.md)

I do have one todo left near the end. I wasn't sure if it was worth it to add an example of Alan trying to implement cancellation correctly or whether leaving it undone would show that it might be out of Alan's reach initially.

Also, should I add more C# examples earlier on to show where Alan is coming from?

### DIFF
--- a/src/vision/status_quo/alan_misses_c_sharp_async.md
+++ b/src/vision/status_quo/alan_misses_c_sharp_async.md
@@ -1,0 +1,414 @@
+# 😱 Status quo stories: Alan misses C# async
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async
+Rust programmers face today. 
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR
+making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories
+[cannot be wrong], only inaccurate). Alternatively, you may wish to
+[add your own status quo story][htvsq]!
+
+## The story
+
+### First attempt
+
+[Alan] has finally gotten comfortable working in rust and finally decides to try writing async code.
+He's used C#'s async and mostly loved the experience, so he decides to try writing it the same way:
+
+```rust
+async fn run_async() {
+    println!("Hello async!");
+}
+
+fn main() {
+    run_async();
+}
+```
+
+But the compiler didn't like this:
+```
+warning: unused implementer of `Future` that must be used
+ --> src/main.rs:6:5
+  |
+6 |     run_async();
+  |     ^^^^^^^^^^^^
+  |
+  = note: `#[warn(unused_must_use)]` on by default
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/alan_misses_c_sharp_async.md
```diff
@@ -0,0 +1,342 @@
...
+### **How would this story have played out differently for the other characters?**
+* [Barbara] may already be used to the ideosynchracies of async in rust. She may not realize how
+difficult it could be for someone who has a very different model of async engrained into them.
+* [Grace] has likely never used async utilities similar to the ones in C# and other GC languages. C
+and C++ tend to use callbacks to manage async workflows. She may have been following the C++
+proposals for coroutines (e.g. `co_await`, `co_yield`, `co_return`), but similar to rust, the
+utilities are not yet thoroughly built out in those spaces. She may be familiar with cancelation in
+external libraries like [`cppcoro`](https://github.com/lewissbaker/cppcoro#Cancellation), or async in
+general with [`continuable`](https://github.com/Naios/continuable)
+* [Niklaus] may not have had enough experience to be wary of some of the pitfalls encountered here.
```
Nashenas88, line 332: I also wasn't sure if this was helpful to have here. I wouldn't expect someone new to programming to be familiar with these pitfalls, but is this too negative of a sentiment to have here?

jbr:
It might be interesting to hear how [stop-token](https://docs.rs/stop-token/0.2.0/stop_token/) (or, very similarly [stopper](https://docs.rs/stopper/0.2.0/stopper/)) meet the c# expectations of what a stop token does

Nashenas88:
I've done a quick review and they look very promising. I'll extend the cancellation section with notes on those. I was surprised that none of those came up in my searches. I discovered that only the tokio one comes up lower in the search when I search `rust async cancellation`, but none of those show up when searching `rust cancel future` or `rust cancel async`.

nikomatsakis (changes requested):
I think the morals could be made more precise! It feels like there are a number of sections here, each of which has a relatively precise moral. Can you note those down? 

For example, one thing I took away was "trying to use things from stdlib is obvious thing to try, but wrong because they are blocking"

Nashenas88:
Sorry for amending the commit. I've been working in gerrit too much lately...

I expanded the cancellable section with an example usage of one of the libaries. I wasn't sure whether to extend it with the painful aspects of discovering `futures::select!` for the first time or not since that's not part of std.

I also fixed some typos in the C# cancellation example and expanded on the morals. Let me know if they're still too vague.

tmandry:
Ecosystem crates are definitely part of the status quo. But you might be able to link to [another story](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_gets_burned_by_select.html) that mentions `select!`.

tmandry (approved):
Update looks good!

nikomatsakis:
Thanks @Nashenas88 !

PR #215, opened 2021-06-02
Harmonic synthesis: skill tree

nikomatsakis:
This is a rough draft for a "skill tree" and some notes about the harmonic synthesis of the work @tmandry and I have done so far.

What this is...

* a set of goals, describing the experience we are shooting for
* a set of technical challenges that will be key parts of that
* a skill showing the progression that leads to that final experiences, and some of the items needed along the way

What this is not...

* a final document: this is a rough draft, suitable for gathering some feedback
* a complete narrative: there will eventually be "shiny future" stories for each of the major items, but we would welcome people writing their own now, or exploring what that might look like
* a detailed technical document: we need to explore the means for implementing those technical challenges in more detail

[Rendered preview.](https://nikomatsakis.github.io/wg-async-foundations/vision/harmonic.html)


jbr:
Is there a key/legend for the various symbols in the skill tree? eg ✋ vs 🛑

nikomatsakis:
Hmm, yes, I should explain that. Probably I should remove the ✋ vs 🛑 actually because they're not really up to date.

The boxes were meant to indicate how much uncertainty there is for each item. Like, how clear of a picture do we have of what this will look like?

nikomatsakis:
I can add a key though

jbr:
This is all super exciting! Is there a shiny future story for "hierarchical structure for tasks with cancelation propagation?" That starts to sound vaguely like an OTP-style supervisor tree, which would be really amazing. Is there more to read somewhere?

nikomatsakis:
@jbr not yet, but that's what I'd like to be working on next. That said, I think that the [Swift structured concurrency](https://github.com/apple/swift-evolution/blob/main/proposals/0304-structured-concurrency.md) proposals or the support in Kotlin is "loosely" what I've been thinking about. There are some technical challenges to make it work, of course.

nikomatsakis:
I think I might remove the "specific technical challenges" section. I am really trying to keep things focused on the *experiences* this far and go step-by-step down to the actual technical challenges that will have to be resolved.

nikomatsakis:
(I will say that I've also been thnking some about Erlang supervisor trees: probably the details of that stuff will play out more in the runtime level, and hence in the ecosystem at large; I think the role of Rust here is to enable the building blocks needed to make it work.)

pickfire:
I think on the top left of the svg we should put some text to say that it is movable. On small screen or some devices, you can only see a white area. Which is not obvious that it is a diagram and is scrollable.

nikomatsakis:
I can do that-- I can also tweak the initial zoom etc, which is pretty grody right now.

nikomatsakis:
I've got another take on this. Closing for now. =)

PR #216, opened 2021-06-03
Add story for library Send-ness.

### LINKED ISSUE #212: Libraries trend towards thread-safety, even if users don't
(Note: I'm not quite sure how to express this as a story without context like "you might want a singlethreaded executor" that doesn't seem super common in the other stories. Advice on framing appreciated!)

### Brief summary

TLDR, if implementing shared library code which of these two should you pick?

```rust
struct Foo {
    cached_subfoos: Arc<Mutex<FooStore>>,
}

// ... or ...

struct Foo {
    cached_subfoos: Rc<RefCell<FooStore>>,
}
```

* implementing library with shared state and/or interior mutability
* need to choose between Rc/Arc and RefCell/Mutex for sharing & mutability and spawning local or Send tasks for async work
* this choice affects the `Send`ness of futures working with the library types
* choosing Rc/RefCell means the library cannot be used in some contexts
* choosing Arc/Mutex is extra overhead in singlethreaded contexts
* offering both "local" and "send" variants of library types is a lot of work if it is feasible at all
* if you only implement one, it's hard to get away with the Rc/RefCell version because it won't work at all in a lot of contexts
* if you implement the Arc/Mutex version and use it in a single-threaded-only context you pay an unnecessary performance overhead
* this problem is amplified if the library accepts Futures in its APIs, as it will likely need to bound those `Future + Send`, forcing single-threaded callers to use Arc etc.

### Optional details

* (Optional) Which [character(s)] would be the best fit and why?
    * [ ] [Alan]: the experienced "GC'd language" developer, new to Rust
    * [ ] [Grace]: the systems programming expert, new to Rust
    * [ ] [Niklaus]: new programmer from an unconventional background
    * [x] [Barbara]: the experienced Rust developer
      * an experienced rustacean is more likely to be implementing libraries for others' consumption where performance tradeoffs matter
      * it's reasonable for a less experienced rust developer to default to Send types everywhere
      * other developers would still likely encounter this as a confusing decision point rather than a frustrating boundary on what they can achieve
* (Optional) What are the key points or morals to emphasize?
  * in synchronous Rust this design choice boils down to "can this library be used with concurrency" but async introduces single-threaded concurrency and more nuance to design tradeoffs

### References

Not sure how to capture this but on Fuchsia, something like 99% of Rust code runs in singlethreaded async contexts that could benefit from `!Send` optimization, but the remaining 1% effectively forces our most common shared libraries to be all-`Send`. I can produce a more formal analysis if that would be helpful.

[character(s)]: https://rust-lang.github.io/wg-async-foundations/vision/characters.html
[project(s)]: https://rust-lang.github.io/wg-async-foundations/vision/projects.html
[Alan]: https://rust-lang.github.io/wg-async-foundations/vision/characters/alan.html
[Grace]: https://rust-lang.github.io/wg-async-foundations/vision/characters/grace.html
[Niklaus]: https://rust-lang.github.io/wg-async-foundations/vision/characters/niklaus.html
[Barbara]: https://rust-lang.github.io/wg-async-foundations/vision/characters/barbara.html

anp:
Hopefully I can sneak this in while we're still in the "winding down" phase for the status quo!

Closes #212.

### DIFF
--- a/src/vision/status_quo/barbara_wants_single_threaded_optimizations.md
+++ b/src/vision/status_quo/barbara_wants_single_threaded_optimizations.md
@@ -1,0 +1,126 @@
+# 😱 Status quo stories: Barbara wants single threaded optimizations, but not that much
+
+## 🚧 Warning: Draft status 🚧
+
+This is a draft "status quo" story submitted as part of the brainstorming period. It is derived from real-life experiences of actual Rust users and is meant to reflect some of the challenges that Async Rust programmers face today.
+
+If you would like to expand on this story, or adjust the answers to the FAQ, feel free to open a PR making edits (but keep in mind that, as they reflect peoples' experiences, status quo stories [cannot be wrong], only inaccurate). Alternatively, you may wish to [add your own status quo story][htvsq]!
+
+## The story
+
+Barbara is working on operating system services, all of which benefit from concurrency, but only some of which benefit from parallelism. In cases where a service does not benefit from parallelism, a single-threaded executor is used which allows spawning `!Send` tasks.
+
+Barbara has developed a useful feature as a module within one of her system's single-threaded services. The feature allows for the creation of multiple IPC objects to use within concurrent tasks while caching and reusing some of the heavier computation performed. This is implemented with reference counted interior mutability:
+
+```rust
+pub struct IpcHandle {
+    cache_storage: Rc<RefCell<IpcCache>>,
+    // ...
+}
+
+struct IpcCache { /* ... */ }
+```
+
+A colleague asks Barbara if she'd be interested in making this code available to other services with similar needs. After Barbara factors the module out into its own crate, her colleague tries integrating it into their service. This fails because the second service needs to hold `IpcHandle`s across yieldpoints and it uses a multi-threaded executor. The multi-threaded executor requires that all tasks implement `Send` so they can be migrated between threads for work stealing scheduling.
+
+### Rejected: both single- and multi-threaded versions
+
+Barbara considers her options to make the crate usable by the multi-threaded system service. She decides against making `IpcHandle` available in both single-threaded and multi-threaded versions. To do this generically would require a lot of boilerplate. For example, it would require manually duplicating APIs which would need to have a `Send` bound in the multi-threaded case:
+
+```rust
+impl LocalIpcHandle {
+    fn spawn_on_reply<F: Future + 'static>(&mut self, to_spawn: impl Fn(IpcReply) -> F) {
+        // ...
+    }
+}
+
+impl SendIpcHandle {
+    fn spawn_on_reply<F: Future + Send + 'static>(&mut self, to_spawn: impl Fn(IpcReply) -> F) {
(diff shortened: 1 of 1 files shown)


src/vision/status_quo/barbara_wants_single_threaded_optimizations.md
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl ClientBroker { /* nothing changed */ }
+```
+
+This incurs some performance overhead when cloning the `Arc` and when accessing the `Mutex`. The former is cheap when uncontended on x86 but will have different performance characteristics on e.g. ARM platforms. The latter's overhead varies depending on the kind of `Mutex` used, e.g. an uncontended `parking_lot::Mutex` may only need a few atomic instructions to acquire it. Acquiring many platforms' `std::sync::Mutex` is much more expensive than a few atomics. This overhead is usually not very high, but it does pollute shared resources like cache lines and is multiplied by the number of single-threaded services which make use of such a library.
+
+## 🤔 Frequently Asked Questions
+
+### **What are the morals of the story?**
```
betamos, line 92: To really drive this point home, maybe finish with that the status quo is using thread-safe data structures in the vast majority of cases, since the added maintenance of dual implementations isn't feasible? Paying for things you don't use (memory barriers, atomics and kernel locks) is arguably a violation of the zero-cost abstraction principle.

src/vision/status_quo/barbara_wants_single_threaded_optimizations.md
```diff
@@ -0,0 +1,126 @@
...
+Abstracting over a library's `Send`ness requires being generic over storage/synchronization types *and* APIs which need to accept user-defined types/tasks/callbacks.
+
+### **What are the sources for this story?**
+
+As of writing, the [Fuchsia](https://fuchsia.dev) operating system had [over 1,500 invocations][st-invocations] of `LocalExecutor::run_singlethreaded`. There were [less than 500 invocations][mt-invocations] of `SendExecutor::run`.[^fuchsia-methods] As of writing the author could not find any widely used support libraries which were not thread-safe.
+
+[st-invocations]: https://cs.opensource.google/search?q=file:rs%20run_singlethreaded&sq=&ss=fuchsia%2Ffuchsia
+[mt-invocations]: https://cs.opensource.google/search?q=file:rs%20%5C.run%5C(&ss=fuchsia%2Ffuchsia
+
+`actix-rt`'s [spawn function](https://docs.rs/actix-rt/1.1.1/actix_rt/fn.spawn.html) does not require `Send` for its futures, because each task is polled on the thread that spawned it. However it is very common when using `actix-rt` via `actix-web` to make use of async crates originally designed for `tokio`, whose [spawn function](https://docs.rs/tokio/1.6.1/tokio/fn.spawn.html) does require `Send`.
```
betamos, line 105: Again, to drive the point home maybe finish with "...which results in applications paying for thread safety that they don't use"?

nikomatsakis (approved):

PR #220, opened 2021-06-28
Add newly merged stories

doc-jones:
As agreed to merge during meeting 2021-06-25 add these stories. Alphabetic arrangement

Alan misses C# async
Barbara begets backpressure and benchmarks async_trait
Barbara polls a Mutex
Barbara wants single threaded optimizations


tmandry:
Thanks!

PR #222, opened 2021-07-09
Harmonic Synthesis, Alpha

nikomatsakis:
This PR represents the **ALPHA RELEASE** of the vision doc:

* The tenets have been replaced with a list of [desired experiences](https://nikomatsakis.github.io/wg-async-foundations/vision/how_it_feels.html). 
* There are no combined narratives yet, but the [roadmap](https://nikomatsakis.github.io/wg-async-foundations/vision/roadmap.html) is (I believe) complete and readable, and the [User's manual of the future](https://nikomatsakis.github.io/wg-async-foundations/vision/shiny_future/users_manual.html) describes the plans in a form that can hopefully give you a good feeling for its impact. There are also clearly enumerated [deliverables](https://nikomatsakis.github.io/wg-async-foundations/vision/deliverables.html), each describing the intended impact as well as design notes.
* The submitted stories have been moved to an [appendix](https://nikomatsakis.github.io/wg-async-foundations/vision/submitted_stories.html). 

### What comes next

* [ ] Continue updating roadmap, design notes, and other details
* [ ] Author combined status quo, shiny future narratives
* [ ] Evaluate impact of proposal on the various projects
* [ ] Review plans with appropriate teams
* [ ] Select owners for the various initiatives and get things going

### How to give feedback

I welcome feedback, but I also don't want this to become a "mega RFC thread for everything ever". Therefore, I request the following:

* Please leave structured reviews, not random comments.
* Feel free to respond to points on other people's reviews, but keep the back and forth to a minimum.
    * I plan to schedule 1:1 time with major stakeholders to go over particular points of interest, so what I'm looking for here is more of a 'heads up, this is an area to dig more into' than a 'let's figure it out in this thread'. If you like, you can take something to Zulip for more conversation (`#wg-async-foundatons` stream).
* If you think I've made factual inaccuracies, or overlooked concerns, I would particularly like to know that. My goal is to turn every major point into a FAQ or an alternative or something like that, so that this document really captures the conversation we have here and elsewhere.


src/vision/harmonic/portable_generic_code/read_and_write.md
```diff
@@ -0,0 +1,155 @@
...
+
+#[repr(inline_async)]
+trait AsyncWrite {
+}
+
+#[repr(inline_async)]
+trait AsyncBidirectional: AsyncRead + AsyncWrite {
+    async fn split(&mut self) -> (impl AsyncRead + '_, impl AsyncWrite + '_)
+}
+```
```
yoshuawuyts, line 155: A "variant D" to consider here would be to implement `(Async)Read + (Async)Write for &T` like `std` ([example](https://doc.rust-lang.org/std/fs/struct.File.html#impl-Read-1)) and `async_std` ([example](https://docs.rs/async-std/1.9.0/async_std/fs/struct.File.html#impl-Read-1)) do. Being able to split into a read and write pair is then as simple as obtaining a shared handle:

```rust
// Using borrows, being entirely zero-cost
let (reader, writer) = &mut (&stream, &stream);
io::copy(reader, writer).await?;

// Creating owned values, using an Arc which implements
// `impl<T> AsyncRead for Arc<T> where &T: AsyncRead {}`
// such as `async_dup::Arc`
let stream = Arc::new(stream);
io::copy(stream.clone(), stream).await?;
```
Because this is a pattern which already exists in the stdlib and has proven to work in async runtimes, I think it's important to include in conversations on how to solve this.

jonhoo:
This is minor, but could your "Rendered" link go directly to https://nikomatsakis.github.io/wg-async-foundations/vision/harmonic.html ? Had to dig a bit through the menu to find it.

src/vision/harmonic/portable_generic_code/read_and_write.md
nikomatsakis: Agreed. That will, however, only work for certain types, right? Not in the general case?

nikomatsakis: I added a placeholder here, I'd like to elaborate on this a bit more.

yoshuawuyts: > That will, however, only work for certain types, right? Not in the general case?

It depends on how one defines "the general case". But we've found this to work for all OS primitives which implement `(Async)Read` and `(Async)Write` (including TCP, unix sockets, and files). And crates such as [`async-h1`](https://docs.rs/async-h1/2.3.2/async_h1/) show protocols can be built on top of this [1].

But yeah, for sure happy to elaborate on this more.

---

[1]: `async_h1::accept` takes `AsyncRead + AsyncWrite + Clone`, which can be any of `TcpStream`, `UnixStream`, or `File` wrapped in an `Arc`. This allows us to work around the lack of async lending iterators in the language right now, at the cost of keeping a refcount.

yoshuawuyts: Also probably worth adding: a key reason _why_ `(Async){Read,Write}` exists on references, is because we (for the most part) cannot guarantee that only a single handle to a file descriptor is open on the system. This is the same rationale why for example [`TcpStream::set_ttl`](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_ttl) takes `&self` rather than `&mut self`.

This also means that `Arc<Mutex<TcpStream>>` and  `Arc<TcpStream>` provide the same correctness guarantees [1], and why `Read + Write for &TcpStream` is as valid as `Read + Write for TcpStream`.

---

[1]: The emphasis is very much on "guarantees" here.

src/vision/harmonic.md
```diff
@@ -0,0 +1,31 @@
+# 🌈 Harmonic synthesis
+
+This page is a **work in progress**. It represents a complete vision for where we want async to go. This vision starts with the [experience] that we want async to provide and then dives into details of the technical pieces that need to get done to create that vision. Note that while a lot of the steps needed are fairly clear, several of them also have [significant unknowns or points of controversy](./harmonic/unresolved_questions.md). We have attempted to highlight those and expect to be working through those points as we go.
```
pickfire, line 3: [experience] link not valid.

pickfire, line 3: I think this could be separated into a **What is this?** part like the rest to be clearer.

src/vision/harmonic/use_current_runtime.md
```diff
@@ -0,0 +1,70 @@
...
+
+The basic idea is to extract out a "core API" of things that a runtime must provide and to make those functions available as part of the `Context` that `Async` values are invoked with. To avoid the need for generics and monomorphization, this would have to be based purely on `dyn` values. This interface ought to be compatible with no-std runtimes as well, which imposes some challenges. Ideally, we would have traits as described in [portable generic code]
+
+## Frequently asked questions
+
+### What about cap-std?
+
+It's interesting to observe that the `dyn` approach is feeling very close to [cap-std](https://blog.sunfishcode.online/introducing-cap-std/). That might be worth taking into consideration. Some targets, like wasm, may well prefer if we took a more "capability oriented" approach.
+
+[portable generic code]: ../portable_generic_code.md
```
eminence, line 66: Broken link

```suggestion
[portable generic code]: portable_generic_code.md
```

eminence (commented):
There's a lot here!  This is an impressive bit of writing.

I didn't read every word, but I skimmed a lot of it.  The overall structure looks fine.  

A quick comment on the "Narrative" section -- I think it would eventually be good to add a hyperlink from each of the narrative subsections to either an anchor in the "Roadmap", or some tracking issue.   Something for a reader to click on after they read a subsection and think to themselves "oh, that sounds neat, where can I learn more or how can I help"

nikomatsakis:
**OK!** I've updated the OP to reflect this, but this PR now represents the "alpha release" of the vision doc:

This PR represents the **ALPHA RELEASE** of the vision doc:

* The tenets have been replaced with a list of [desired experiences](https://nikomatsakis.github.io/wg-async-foundations/vision/how_it_feels.html). 
* There are no combined narratives yet, but the [roadmap](https://nikomatsakis.github.io/wg-async-foundations/vision/roadmap.html) is (I believe) complete and readable, and the [User's manual of the future](https://nikomatsakis.github.io/wg-async-foundations/vision/shiny_future/users_manual.html) describes the plans in a form that can hopefully give you a good feeling for its impact. There are also clearly enumerated [deliverables](https://nikomatsakis.github.io/wg-async-foundations/vision/deliverables.html), each describing the intended impact as well as design notes.
* The submitted stories have been moved to an [appendix](https://nikomatsakis.github.io/wg-async-foundations/vision/submitted_stories.html).

src/vision/deliverables/async_fn/async_fn_in_traits.md
```diff
@@ -0,0 +1,53 @@
...
+## Frequently asked questions
+
+### Can users easily bound those GATs with `Send`, maybe even in the trait definition?
+
+- People are likely to want to say "I want every future produced by this trait to be Send", and right now that is quite tedious.
+- This applies equally to other "`-> impl Trait` in trait" scenarios.
+
+### What about "dyn" traits?
+
+- See the sections on "inline" and dyn" async fn in traits below!
```
MarcoIeni, line 53:
```suggestion
- See the sections on "inline" and "dyn" async fn in traits below!
```
?

src/vision/deliverables/async_fn/async_drop.md
```diff
@@ -0,0 +1,115 @@
...
+Within async functions, when we drop a value, we will invoke "async drop glue" instead of "drop glue". "Async drop glue" works in the same basic way as "drop glue", except that it invokes `AsyncDrop` where appropriate (and may suspend):
+
+* The async drop glue for a type T first executes the `AsyncDrop` method
+    * If `T` has no `AsyncDrop` impl, then the glue executes the synchronous `Drop` impl
+        * If `T` has no `Drop` impl, then this is a no-op
+* The async drop glue then recursively "async drops" all fields of T
+
+### Requires inline fn 
+
+Making this work requires [inline async fn]. This is because Rust presently assumes *all* types are croppable. Consider a function `foo`:
```
tmandry, line 38:
```suggestion
Making this work requires [inline async fn]. This is because Rust presently assumes *all* types are droppable. Consider a function `foo`:
```

src/vision/deliverables/async_fn/async_drop.md
```diff
@@ -0,0 +1,115 @@
...
+
+The best solution here is unclear. We could have an "allow-by-default" lint encouraging explicit use of `async_drop`, but as the code above shows, the result may be highly unergonomic (also, imagine how it looks as the number of variables requiring async-drop grows).
+
+Another option is to target the problem from another angle, for example by adding lints to identify when large values are stored in a future or on the stack, or to allow developers to tag local variables that they expect to be stored on the stack, and have the compiler warn them if this turns out to not be true.
```
tmandry, line 95: Does the second idea imply some kind of implicit allocation?

src/vision/deliverables/async_fn/async_drop.md
```diff
@@ -0,0 +1,115 @@
...
+
+It is easy enough to make async-drop be used, but it is currently not possible to prevent sync drop, even from within an async setting. Consider an example such as the following:
+
+```rust
+async fn foo(dbc: DatabaseConnection) -> io::Result<()> {
+    drop(dbc);
+}
+```
+
+The compiler could however lint against invoking (or defining!) synchronous functions that take ownership of values whose types implement `AsyncDrop`. This would catch code like the case above. We may have to tune the lint to avoid false warnings. Note that it is important to lint *both* invocation *and* definition sites because the synchronous function may be generic (like `drop`, in fact).
```
tmandry, line 107: What do you mean by linting definition sites? Something that takes `impl Async` and drops it?

src/vision/deliverables/async_fn/async_drop.md
```diff
@@ -0,0 +1,115 @@
...
+
+The question remains: what should code that implements `AsyncDrop` *do* if synchronous `Drop` is invoked? One option is panic, but that is suboptimal, as panic from within a destructor is considered bad practice. Another option is to simply abort.
```
tmandry, line 109: Another option is portable `block_on` that actually works, including if called from within an async task. An executor could detect this case and log a warning (or panic/abort).

src/vision/deliverables/async_fn/async_drop.md
```diff
@@ -0,0 +1,115 @@
...
+
+Preventing this 'properly' would require changing fundamental Rust assumptions (e.g., by introducing the `?Drop` trait). While such a change would make Rust more expressive, it also carries complexity and composition hazards, and would require thorough exploration. It is also a step that could be taken later (although it would require some form of explicit opt-in by types to avoid semver breakage).
```
tmandry, line 111:
```suggestion
Preventing this 'properly' would require changing fundamental Rust assumptions (e.g., by introducing the `?Drop` trait). While such a change would make Rust more expressive, it also carries complexity and composition hazards, and would require thorough exploration. It is also a step that could be taken later (although it would require some form of explicit `impl !Drop` opt-in by types to avoid semver breakage).
```

It took me a minute to realize what exactly you meant.

nrc:
Reading through this as a relative outsider, I think there is a strong vision for the end goal and the roadmap/deliverables seem pretty exhaustive, but what I'm missing is a description of how the deliverables will satisfy the goals in the vision. IOW, even if everything in the roadmap gets ticked off, I can't figure out if we will have arrived at the shiny future or not.

src/vision/deliverables/borrowed_data_and_cancellation.md
```diff
@@ -0,0 +1,117 @@
...
+        std::async_iter::from_iter(work)
+            .map(|item| do_select(database, work_from_item(item)))
+            .buffered_in(5, scope)
+            .for_each(|work_item| process_work_item(database, work_item))
+            .await;
+    }).await;
+}
+```
+
+### Concurrency without scopes: Join, select, race, and friends
```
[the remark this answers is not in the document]
nikomatsakis: I thought about that, but I don't think that's correct. I don't think it's correct that we should be concurrently fetching the next item for all async iterators while also processing the current one. It might be what you want sometimes, but I think if so, you should ask for it, which is precisely what I'm proposing here =)

src/vision/deliverables/borrowed_data_and_cancellation.md
```diff
@@ -0,0 +1,117 @@
...
+  - This would mean that typical `async fn` could not be given to select, though we might allow people to mark `async fn` as "cancel-safe", in which case they would implement `Future`. They would also not have access to ordinary async fn, though.
+    - Effectively, the current `Future` trait becomes the "cancel-safe" form of `Async`. This is a bit odd, since it has other distinctions, like using `Pin`, so it might be preferable to use a 'marker trait'.
+  - Of course, users could spawn a task that calls the function and give the handle to select.
+
+## Frequently asked questions
+
+### Could there be a convenient way to access the current scope?
+
+If we wanted to integrate the idea of scopes more deeply, we could have some way to get access to the current scope and reference its lifetime. Let's imagine we added a keyword `scope`, and we said that its lifetime is `'scope`. One would then be able to do something like the following:
+
```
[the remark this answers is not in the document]
nikomatsakis: I'm just going to delete that. After further exploration I didn't find much reason to go this path, and I don't think it would fit that well with Rust. Also, to be honest, I'm not entirely sure I had in mind.

src/vision/deliverables/borrowed_data_and_cancellation/capability.md
```diff
@@ -0,0 +1,23 @@
...
+
+Async functions are commonly written with borrowed references as arguments:
+
+```rust
+async fn do_something(db: &Db) { ... }
+```
+
+but important utilities like `spawn` and `spawn_blocking` require `'static` tasks. Without "unfogettable" traits, the only way to circumvent this is with mechanisms like `FuturesUnordered`, which is then subject to footguns as described in [Barbara battles buffered streams](https://rust-lang.github.io/wg-async-foundations/vision/status_quo/barbara_battles_buffered_streams.html).
+
+There are two main approaches under consideration to address this issue:
```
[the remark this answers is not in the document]
nikomatsakis: Added a note about that

src/vision/deliverables/async_fn/async_drop.md
nikomatsakis: Nothing in this paragraph implies implicit allocation to me -- just warnings when things didn't turn out how you might expect. You can choose how to resolve that.

nikomatsakis: Added a sentence that hopefully clarifies.

src/vision/deliverables/async_fn/async_drop.md
nikomatsakis: Added a note about how this is the "async-sync-async" sandwich problem if another guise.

src/vision/deliverables/async_fn/async_drop.md
nikomatsakis: Updated

nikomatsakis:
Thanks @nrc. In the "User's Guide", there were links to specific deliverables that implement each point, I wonder if part of what was missing was that the "concrete designs" aren't really known yet. We have things in mind but it's somewhat intentionally vague, and instead meant to describe the intended *experience*, so that we have space to work out the best path as we go. (To be included, though, I felt like I had to have "reasonably strong" confidence that the idea can be done.)

nikomatsakis:
Hi all! I've been talking with @tmandry and these are the next steps we have in mind:

* To start, we're going to merge this PR, since it's basically the right starting point (that will require a somewhat annoying rebase...).
* Next, we're going to upgrade the status quo and shiny future stories to have narratives based on a few chosen projects that illustrate the major problems that we see. I have a branch laying out some of that work, but I'm going to open it as a follow-up PR.
* We're actively discussing who to setup as owners of each area and we'll be following up on that.

nrc:
> Thanks @nrc. In the "User's Guide", there were links to specific deliverables that implement each point, I wonder if part of what was missing was that the "concrete designs" aren't really known yet. We have things in mind but it's somewhat intentionally vague, and instead meant to describe the intended experience, so that we have space to work out the best path as we go. (To be included, though, I felt like I had to have "reasonably strong" confidence that the idea can be done.)

Yeah, the links from the user guide were pretty useful, I guess I was looking for maybe a document or section that was just that mapping for quick reference (rather than being embedded in the user guide) and with some explanation of how the deliverable satisfies the goals in 2.2 and 2.6, where it is less obvious (especially 2.2, the mapping to 2.6 is a bit clearer IMO and one also has to be able to map the concrete goals of 2.6 to the abstract goals of 2.2).

PR #238, opened 2021-09-05
Add tool for fixing broken links

tmandry:


eholk (approved):
It seems like it might be worth pulling this tool out into an mdbook plugin.

PR #258, opened 2021-11-10
Add design notes for async stack traces

eholk:
This adds a page describing the current state of async stack traces,
what some of the problems are, and some targetted changes that could
improve them.

Issue #251

### DIFF
--- a/src/design_docs/async_stack_traces.md
+++ b/src/design_docs/async_stack_traces.md
@@ -632,6 +632,10 @@ It does help in the most common cases, however, and when users take a more manua
 
 There are several ways to do this, with varying levels of implementation effort and cost.
 
+As to what this would look like for library authors, ideally we would have something like a `#[backtrace_transparent]` attribute that is applied to a function and indicates that the function should be hidden from backtraces by default.
+
+There are several ways we could do the underlying implementation, which are discussed below.
+
 The conceptually simplest is to allow multiple `__rust_begin_short_backtrace`/`__rust_end_short_backtrace` pairs.
 Implementing this could be done almost entirely with changes to [`_print_fmt`].
 This approach has some serious drawbacks.
@@ -645,6 +649,7 @@ It might be possible instead to include more information in the debugging symbol
 For example, we might be able to add a flag indicating that a certain function should be hidden from the backtrace.
 To do this, we would first need to make sure existing debugging formats such as DWARF and PDB are able to encode such information.
 If there is already support for this, then it's likely debuggers would benefit as well since they would also be able to display trimmed backtraces.
+It is worth noting that this solution would not help much for builds without debug symbols.
 
 A third option is to use some kind of name based heuristics.
 For example, by default we may want to only show frames in the root crate, although this may be too restrictive for large projects.
--- a/src/design_docs/async_stack_traces.md
+++ b/src/design_docs/async_stack_traces.md
@@ -1,0 +1,705 @@
+# Async Stack Trace Design Notes
+
+This page has notes on the state of [async stack traces], highlights specific issues with current stack traces, and suggests changes to improve these issues.
+
+The two main suggestions are:
+
+1. Allow async runtimes to control where the short backtrace cutoff happens
+2. Expand the options allowed in `RUST_BACKTRACE` to support including/excluding frames from certain crates or module paths in the backtrace.
+
+[async stack traces]: ../vision/roadmap/polish/stacktraces.md
+
+## The Current State of Things
+
+The current state of stack traces was captured pretty well in the story [Barbara Trims a Stack Trace][barbara-trims-stack-trace]. We've recreated a similar example to the one in the story here. We'll look at several executors.
+
+[barbara-trims-stack-trace]: https://rust-lang.github.io/wg-async-foundations/vision/submitted_stories/status_quo/barbara_trims_a_stacktrace.html
+
+### Tokio
+
+<details><summary>Short Backtrace</summary>
+
+```
+thread 'main' panicked at 'explicit panic', C:\Users\ericholk\repo\backtrace-examples\async-common\src\lib.rs:10:5
+stack backtrace:
+   0: std::panicking::begin_panic_handler
+            at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35\/library\std\src\panicking.rs:517
+   1: core::panicking::panic_fmt
+            at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35\/library\core\src\panicking.rs:101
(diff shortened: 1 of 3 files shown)


src/design_docs/async_stack_traces.md
```diff
@@ -0,0 +1,705 @@
...
+             at .\src\main.rs:14
+   4: sync::bar
+             at .\src\main.rs:10
+   5: sync::foo
+             at .\src\main.rs:6
+   6: sync::main
+             at .\src\main.rs:2
+   7: core::ops::function::FnOnce::call_once<void (*)(),tuple$<> >
+             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35\library\core\src\ops\function.rs:227
+note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
nrc, line 449: As you can see regular backtraces are also pretty sub-optimal. IMO, frames 0, 1, 2, and 7 should be elided too. In real backtraces, there are often trampoline functions, etc. which should be elided too.

src/design_docs/async_stack_traces.md
```diff
@@ -0,0 +1,705 @@
...
+```
+
+To achieve parity with the synchronous backtrace, there are two more improvements needed.
+
+The first is that in between each frame in user code, there is a frame with a call to `core::future::from_generator`.
+
+The second is that rather than seeing a call to `common::foo` or similar, we see `common::foo::generator$0`.
+
+These two are relatively minor issues.
+Fixing them may not be desirable.
```
nrc, line 612: I disagree, I think that in aggreagate, these issues make async backtraces pretty painful and it is very desirable to fix them :-)

src/design_docs/async_stack_traces.md
```diff
@@ -0,0 +1,705 @@
...
+Then, async runtimes that provide a macro such as `#[tokio::main]` or `#[async_std::main]` would arrange to start up through this alternate path and call `__rust_begin_short_backtrace` shortly before invoking user code.
+
+[`rt.rs`]: https://github.com/rust-lang/rust/blob/master/library/std/src/rt.rs
+
+Note that this approach does not help in cases where users create and launch the runtime manually rather than using a library-provided macro.
+It does help in the most common cases, however, and when users take a more manual approach they would also be able to control where the short stack trace starts.
+
+### Allow trimming of internal frames
+
+There are several ways to do this, with varying levels of implementation effort and cost.
```
nrc, line 633: These are pretty implementation-focussed. My preferred solution is that any function can have an attribute, say `#[backtrace_transparent]`, and then it is elided from the default backtrace. Of course, how this is implemented is a hard question; I think either of your first two would work, but neither feel great.

src/design_docs/async_stack_traces.md
```diff
@@ -0,0 +1,705 @@
...
+First, it requires a lot of work from library authors to annotate each transition point between user and library code.
+Second, `__rust_begin_short_backtrace`/`__rust_end_short_backtrace` are built to defeat inlining to ensure they show up as a frame on the stack.
+This is fine when the functions are only called during process startup and once when a panic starts, but it would likely be prohibitively expensive if interspersed between every async function call.
+
+[`_print_fmt`]: https://github.com/rust-lang/rust/blob/master/library/std/src/sys_common/backtrace.rs#L52
+
+It might be possible instead to include more information in the debugging symbols.
+For example, we might be able to add a flag indicating that a certain function should be hidden from the backtrace.
+To do this, we would first need to make sure existing debugging formats such as DWARF and PDB are able to encode such information.
+If there is already support for this, then it's likely debuggers would benefit as well since they would also be able to display trimmed backtraces.
```
nrc, line 647: This wouldn't work for builds without debug symbols, presumably.

eminence:
Two big thumbs up to efforts around making stack traces more readable.   👍 👍 

This part is interesting to me:

> For example, we might be able to add a flag indicating that a certain function should be hidden from the backtrace. To do this, we would first need to make sure existing debugging formats such as DWARF and PDB are able to encode such information.

because I am currently trying to debug a hang/deadlock.  Since I don't have a crash that goes through the usual rust stack printing machinery, I'm trying to look at this with gdb.  In my case I have 9 threads each with about 50 frames; it's all pretty inscrutable.

I wonder if it would be useful to have some type of standalone web page or other utility into which you could paste a bunch of async stack frames.  The utility would then trim/elide/annotate the frames as necessary to produce something that's easy-to-understand.  This may be more work than just experimenting with `_print_fmt` directly, but it might be a useful way to experiment with different UIs or trimming heuristics.

src/design_docs/async_stack_traces.md
eholk: I like that idea too. I'll add a note describing that idea.

It occurs to me that we could probably experiment with some of these ideas with an external crate before we commit to doing this in `std` or `rustc`.

src/design_docs/async_stack_traces.md
eholk: Good point. I'll add a note making that explicit.

src/design_docs/async_stack_traces.md
eholk: Yeah.. I feel like when I wrote this I thought there was a good reason for doing it this way, but looking at it a month later I'm not sure why.

In the second case (`common::foo::generator$0`), maybe I thought that the user would care that there is a generator involved. On second thought, I don't know why they would. It's not like they wrote a generator, so having the one that rustc generator leak out this way is probably surprising and confusing.

On the first case, I think I was nervous about completely eliding stack frames from the user. Again, it's not clear why this information is useful, but I'd feel a little cheated to find out the debugger was hiding stack frames from me that were actually there. That said, because of inlining, the debugger is probably already telling me something that's not quite accurate.

src/design_docs/async_stack_traces.md
eholk: Yeah, I think whatever we do for async stack traces will end up improving sync backtraces quite a bit too.

eholk:
> I wonder if it would be useful to have some type of standalone web page or other utility into which you could paste a bunch of async stack frames. The utility would then trim/elide/annotate the frames as necessary to produce something that's easy-to-understand. This may be more work than just experimenting with `_print_fmt` directly, but it might be a useful way to experiment with different UIs or trimming heuristics.

@eminence - I think this is a good idea. At the very least, we could put together a library that can filter backtraces that would be easy to write additional tools around. Doing so would let us experiment a lot easier on what backtraces should look like.

eholk:
Does anyone else have comments or questions I should address before merging this? I'll give a little more time for comments but then I'll plan on going ahead and merging it and we can make further updates as needed.

PR #261, opened 2022-01-06
improve information on must_not_suspend page

guswynn:
Adds information about what is blocking the current stabilization, let me know what else I should add!


eholk:
Looks good to me, thanks for the quick turnaround!

PR #264, opened 2022-01-11
[async-stacktraces] Add references to folly blogs

eholk:
Facebook did a five part series of blog posts on async stack traces in [folly]. This is in the context of C++, but it has a lot in common with what we might want for Rust so it's worth keeping a reference here.

[folly]: https://github.com/facebook/folly

Related to #251


yoshuawuyts (approved):
Oh wow, this is great. Thanks for sharing!

PR #339, opened 2024-05-10
update meeting time; collect recent meeting notes

ahl:
The meeting time has changed. I'm not sure when. I scooped up the meeting notes that seemed obvious to find; I'm not sure if I missed some of the notes or if those meetings never happened.

### DIFF
--- a/src/meetings.md
+++ b/src/meetings.md
@@ -17,11 +17,74 @@ See also:
 
 ## Meeting schedule
 
-Our weekly meetings are held at [09:00 PT].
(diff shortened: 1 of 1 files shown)


yoshuawuyts (approved):
This is great; thank you so much for doing this!

The meeting time changed in week 15 (2024-04-11). There was a standing conflict with another meeting, so we moved it to be an hour later. I've pushed an update to your branch with the new times - with that I think this should be good to merge!

PR #340, opened 2024-05-11
fix [10:00] link

ahl:
My last push screwed up the link to the [10:00] time. This should address that.


vincenzopalazzo (approved):
Thanks!

